PR #5, opened 2020-04-17
initial support for systemd

rphillips:
Adds initial support for per profile systemd user files where a user can run `resticprofile -n configs systemd-unit daily`. This is super useful to allow for automation to 'install' everything needed to automate backups.

Adds:

```
resticprofile -n [profile] systemd-unit [OnCalendar setting]
```

The following snippet runs resticprofile then enables the systemd timer. The intention is to create 'user' systemd profiles. If you think this PR will get merged, the next step would be add system wide support.

```
resticprofile -n configs systemd-unit daily
systemctl --user enable resticprofile-backup@configs.timer
```

I added the xdg library to pull in ~/.config/resticprofile/profile.conf on linux (the library works on windows and OSX as well)... The entire logic in filesearch/filesearch.go could be simplified with the xdg library, but I left it as is to maintain backwards compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -660,3 +660,61 @@ stdin:
         - stdin
 
 ```
+
+## Using resticprofile and systemd
+
+systemd is a common service manager in use by many Linux distributions.
+resticprofile has the ability to autocreate systemd timer and service files.
+systemd can be used in place of cron to schedule backups.
+
+All systemd units are created under the user's systemd profile (~/.config/systemd/user).
+
+TODO: create system profiles
+
+### systemd calendars
+
+resticprofile uses systemd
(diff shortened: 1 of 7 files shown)


creativeprojects:
This is interesting indeed

I'm still old school and add lines into a crontab, but that's a good idea.

Next on the agenda would be to add a launchd task in Mac OS X and possibly the same with Task Manager on Windows (yes, I do use all three)

creativeprojects:
Hey!

So I had a look at it in more details, and yes that's quite simple. I like the idea of it.

- Is there an easy way to validate the `OnCalendar` value before generating the file? systemd will probably complain if it's wrong
- Could we run the `systemctl --user enable` automatically? either with a `--enable` flag or maybe the other way around, without a `--no-enable` flag?

PR #16, opened 2020-11-08
Escape `exclude` globs passed to `/sh`

### LINKED ISSUE #15: Wildcard path expansion leads to excludes turning into additional source paths
### Problem Description

I have a `backup` config section using `iexclude` and ran into an issue where wildcard paths are expanded and lead to what should be excluded paths turning into additional backup paths.

Example config:
`iexclude = ["bin", "src/**/node_modules", "go/pkg"]`

When resticprofile runs it builds up this command to invoke restic: `sh -c  restic backup --iexclude bin --iexclude src/**/node_modules --iexclude go/pkg  --tag home /home/kevin`

So far so good.  However, when this command is shelled out and executed, the wildcards are expanded by `sh` into this command: `restic backup --iexclude bin --iexclude src/app1/node_modules src/app2/node_modules src/appN/node_modules --iexclude go/pkg <remaining args> <source_path>`

Note that the expanded paths following first found path (`src/app1/node_modules`) are NOT preceded by `--iexclude`.  This causes restic to treat `src/app2/node_modules`, etc as additional sources to include in the backup:

```
ID        Time                 Host        Tags        Paths
------------------------------------------------------------------------------------------------------------------------------
ffc66e1f  2020-09-06 09:04:35  desktop     home        /home/kevin
                                                       /home/kevin/src/app2/node_modules
<SNIP>
```

### Workaround and Possible Solutions

To workaround this, you can escape the `*` chars in the config so that `sh` does not expand them itself and lets restic handle globbing: `"src/\\*\\*/node_modules"`

A couple ideas to fix this:
- Automatically escape the `*` chars, similar to how other paths are handled via `fixPath` (I note that `iexclude` and the like fall into the `OtherFlags` list and are not "well known" as paths in the current code, so these may need to be made into "first class" config keys and handled as paths).
- Make a note in the docs that this escaping is necessary.
- [insert your better idea here]

kppullin:
This prevents `sh` expansion of any globs set by `exclude` & `iexclude` and allows `restic` to apply its own globbing logic.

Fixes #15.

### DIFF
--- a/config/path.go
+++ b/config/path.go
@@ -46,6 +46,13 @@ func unixSpaces(value string) string {
 	return value
 }
 
+func unixGlobs(value string) string {
+	if runtime.GOOS != "windows" {
+		value = strings.ReplaceAll(value, "*", `\*`)
+	}
+	return value
+}
+
 func absolutePrefix(prefix string) pathFix {
 	return func(value string) string {
 		if value == "" ||
--- a/config/path_test.go
+++ b/config/path_test.go
@@ -24,13 +24,14 @@ func TestFixUnixPaths(t *testing.T) {
 		{"~/dir", "~/dir"},
 		{"$TEMP_TEST_DIR/dir", "/home/dir"},
 		{"some file.txt", "prefix/some\\ file.txt"},
+		{"/**/.git", "/\\*\\*/.git"},
 	}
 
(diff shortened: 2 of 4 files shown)


creativeprojects:
It looks good to me, thanks 👍🏻

PR #19, opened 2020-12-13
Add background types and low priorityIO to launchd plist

EkilDeew:
I've hard coded the entries for now. I would have liked to change the background types and the IO key according to resticprofile global config but I don't want to mess up your coding style nor your structure.
So i'm waiting for your insight on this.

### DIFF
--- a/config/profile.go
+++ b/config/profile.go
@@ -232,6 +232,7 @@ func (p *Profile) Schedules() []*ScheduleConfig {
 				environment: p.Environment,
 				nice:        10, // hard-coded for now
 				logfile:     s.ScheduleLog,
+				priority:    s.SchedulePriority,
 			}
 
 			configs = append(configs, config)
--- a/config/schedule.go
+++ b/config/schedule.go
@@ -12,6 +12,7 @@ type ScheduleConfig struct {
 	jobDescription   string
 	timerDescription string
 	nice             int
+	priority         string
 	logfile          string
 }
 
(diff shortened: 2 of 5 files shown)


creativeprojects:
Sorry I haven't got back to you yet.
You're right, we cannot use the global configuration for that. We can keep the command line configuration and the scheduling configuration separate.

I want to implement something that could be reused in the other systems (systemd and Windows task scheduler).

We could add an option `schedule-priority` with two options:
- `background`
- `standard`

Then:
- on Mac, `background` would automatically set `LowPriorityIO` to true, there's no need for a separate flag for it.
- on Linux it will set the nice and ionice fields
- on Windows I might need to set the priority manually, like it is done right now. But we can still keep a separate flag for it.

How does that sound?

EkilDeew:
Sounds good however i've only successfully added the keys for darwin schedule, I have no experience in systemd schedule nor Windows.
What i'd suggest is that you add `schedule-priority` flag with the options and manage the linux and Windows one, and i'll append Darwin schedule configuration.

I'm up to do it myself if you don't have time to commit, i'll just take a bit more time 😀

creativeprojects:
I'm very sorry it's taking so long.
So in https://github.com/creativeprojects/resticprofile/commit/a8939edb097d97034c24c7e913ec5faa5b0733b3 I've added a parameter `schedule-priority` and two constants `background` and `standard`: You can use them to generate the values of ProcessType and LowPriorityIO. (you just need to rebase your PR from master)

In the meantime I'll look into using this parameter with systemd and windows task manager 👍🏻

EkilDeew:
Don't be sorry ! Not everyone has time to work on open source :).
I've rebased and implemented priority for the darwin plist based on `schedule-priority` parameter. I've added a test as-well.

I'm not used to Go, so i'm all ears to critics !

creativeprojects:
Cool, it works nicely on Mac. I'm accepting the PR and I'll update the code for Linux and Windows,
Thanks 👍🏻

Issue #23, opened 2020-12-20
Scheduling "retention" reads policy from "forget"


jkellerer:
### Finding

At the moment the following configuration schedules `forget` without any effect (at least with `systemd`):
```yaml
default:
  retention:
    prune: true
    keep-last: 14
    schedule: '@daily'
    schedule-permission: 'system'
```

The reason is that `retention` isn't a command but a configuration section that is mapped to `forget`. The schedule invokes `forget` which reads the retention config from the `forget` config section (that is what I'd expect when `forget` is scheduled). 
With the following workaround, it is possible to schedule `forget` (forget itself cannot be scheduled directly):
```yaml
default:
  forget:
    prune: true
    keep-last: 14
  retention:
    schedule: '@daily'
    schedule-permission: 'system'
```

### Suggestion

The idea to have `retention` policies defined with `backup` is great since it allows to have separate retention settings per tag & path. When called before or after backup and without `prune` (for large repos) it is also a fast operation. So this should not be changed, it is good as it is.

However, instead of artificially mapping `retention` to a scheduled `forget` command, I'd suggest to allow scheduling `forget` directly and remove the option to schedule `retention` (it is broken anyway).

See also #22 , when prune can be scheduled separately, then there are less use cases where scheduling `forget` (with or without prune) really makes sense, still I'd allow it as there may be cases where it is needed.

jkellerer:
(I'm happy to provide a PR if this is accepted)

creativeprojects:
hmmm yeah, I suppose being able to schedule `forget` makes sense, probably a bit more than `retention` which is supposed to be running along with a backup.

My intention was to keep only one configuration when running with a backup, or when running separately. But thinking about it, if you need to run it separately it means you probably need a different configuration anyway.

jkellerer:
Yes, that’s also how I see it. Running along with backup is where retention makes perfectly sense. And in most cases that’s where it should run. 

Main point for me is to run prune separately. But for the remaining cases where forget may need to be scheduled, it likely would need its own config as you also mentioned. 

So instead of fixing the bug it might be better to adjust the logic as proposed above.

creativeprojects:
> So instead of fixing the bug it might be better to adjust the logic as proposed above.

I agree we can add a schedule of the `forget` command.
But I wouldn't want to break anybody's configuration relying on the `retention` schedule: we could fix the issue, or at a minimum I would mark it deprecated and then remove it later.

jkellerer:
[...]

PR #24, opened 2020-12-20
Added prune as supported, schedulable command

### LINKED ISSUE #22: Allow to schedule "prune"
This is a request to add `prune` to the list of `ScheduledSections`.

At the moment `check`, `backup` and `retention` can be scheduled. However one of the heaviest tasks, `prune` is not available. While retention can do `prune` as well it isn't very practical for large repositories (e.g. in my case, it runs +1 hour for a 500G repo on ordinary HDDs). 

While retention policies can differ per path and tag, `prune` should just be run once per-repo (and at a lower rate). A separate schedule option would allow to configure this easily.

jkellerer:
Fixes #22 

Implementation Notes:
* `profiles_test` & `show` was checked in with forced CRLF (other files were not). 
  The diff (without ignored whitespace) is larger than it should. Needs to be clarified if CRLF should be forced.
* I've consolidated schedule options since #23 may need one more schedulable section.
* All schedulable sections print with one newline in `show`. Needs to be clarified if this should be fixed.

~~Marked as WIP since I only tested it via `profiles_test` and on my mac so far.~~

Tested in regular backup schedule and via unit and manual tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1770,6 +1770,14 @@ Flags passed to the restic command line
 * **read-data-subset**: string
 * **with-cache**: true / false
 
+`[profile.prune]`
+
+Flags used by resticprofile only
+
+* **schedule**: string OR list of strings
+* **schedule-permission**: string (`user` or `system`)
+* **schedule-log**: string
+
 `[profile.mount]`
 
 Flags passed to the restic command line
--- a/config/profile.go
+++ b/config/profile.go
@@ -1,6 +1,8 @@
 package config
 
 import (
+	"reflect"
+
 	"github.com/creativeprojects/clog"
 	"github.com/creativeprojects/resticprofile/constants"
 )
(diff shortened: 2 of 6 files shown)


jkellerer:
Self note: `resticprofile show` returns invalid content with this change. Needs to be checked.

jkellerer:
Please note that the fix for `show` adds one newline. I‘d fix this before merge, however would like to get feedback first.

creativeprojects:
Sure, I have been a bit busy, sorry about that.
I'll have a look at it in the next few days 👍🏻 

Thanks

creativeprojects:
I see the two annoying files with CRLF ending, it's because sometimes I fix things on my work laptop.
To make things easier I can change them to LF and then maybe we can rebase the PR?

Sorry about that

jkellerer:
Yes sure, that would be great.

creativeprojects:
Can you try a rebase now?

creativeprojects:
Cool, thanks

I like the way you refactored the scheduling config into `ScheduleSection`, nice one 👍🏻 

I understand now why you had to change the code of the `show` command, I have to admit I was a bit puzzled over the holidays when I glanced at the PR message 😆 

I'm looking at the newline issue; I realise the code of the `show` command is rather messy. I remember I was learning go reflection on this one...

creativeprojects:
It's ok for the `show` command, I need to work on it later.

Thanks fo the PR 👍🏻

jkellerer:
Thanks a lot for reviewing and merging!

PR #26, opened 2021-01-09
Add support for scheduling forget

jkellerer:
As discussed in #23 , this adds support for scheduling forget.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1737,6 +1737,12 @@ Flags passed to the restic command line
 
 `[profile.forget]`
 
+Flags used by resticprofile only
+
+* **schedule**: string OR list of strings
+* **schedule-permission**: string (`user` or `system`)
+* **schedule-log**: string
+
 Flags passed to the restic command line
 
 * **keep-last**: integer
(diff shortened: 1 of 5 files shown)


jkellerer:
Let me know, if tests are required for the 2 refactored lines

creativeprojects:
> Let me know, if tests are required for the 2 refactored lines

I wouldn't say required, but if they are easy to test, that would be a nice to have.

Otherwise that looks good to me 👍🏻

jkellerer:
Well, I guess if code analysis already tells us that we shouldn't be lazy, we should listen to it :)

I've kept the original test and leave it up to you if both tests should remain. The original test still has some extra value since it covers 2 conditions (true and actual value) in a single invocation.

jkellerer:
Not entirely certain what caused the test aborts. Tests run fine locally (besides TestUpdate which was very slow today, but is skipped here).

creativeprojects:
> Not entirely certain what caused the test aborts. Tests run fine locally (besides TestUpdate which was very slow today, but is skipped here).

I have a flaky test since I moved from Travis CI to GitHub Actions. Not sure what's happening, it aways works fine locally (TestInterruptShellCommand)

creativeprojects:
Nice work. I'll gradually update my backup configurations as I have a few using `retention` and I'll display a deprecation warning.

Thanks

Issue #27, opened 2021-01-09
Add option "--all" to schedule/unschedule commands


jkellerer:
I was wondering whether it would be possible to add the CLI param "--all" to schedule and unschedule in order to remove and update schedules of all profiles. 

Currently I'm using a shell script that lists profile names and calls un/schedule on all of them. But since I cannot know if schedules exist, I cannot perform any error handling in the shell script. A built-in option with proper error handling would be better.

Also one thing I found is that "unschedule" only operates on defined schedules. However in reality you modify the config and then think, lets test the changes now. This may leave schedules behind since only defined schedules will be removed. Remove should try to remove all possible schedules and silently ignore errors on "ErrorServiceNotFound" for schedules that are not defined.

As always, I'm happy to support this with a PR.

creativeprojects:
> Also one thing I found is that "unschedule" only operates on defined schedules. However in reality you modify the config and then think, lets test the changes now. This may leave schedules behind since only defined schedules will be removed. Remove should try to remove all possible schedules and silently ignore errors on "ErrorServiceNotFound" for schedules that are not defined.

This is a very good point, and I thought about it before. I have written down this command on my notes for later:

`systemctl list-units --all --no-legend "resticprofile-*@profile-*.timer"`

which should give us a full list of schedules

Or maybe a more specific:

`systemctl list-units --all --no-legend --type timer "resticprofile-*@profile-*"`

As you can see I haven't got the time to do it yet, so if you fancy having a go at a `schedule --all` and `unschedule --all` I think that would be great 👍🏻

jkellerer:
Good to know.

Using `systemctl list-units --all --no-legend --type timer "resticprofile-*@profile-*"` is a nice optimisation but also requires changes to the interface and implementations of all supported schedulers. Was thinking to implement it with existing interfaces first since that would probably be enough.

We can discuss this on the PR (when ready).

jkellerer:
Let me know what you think about #28 , is it ok to continue that way? 

(I'd implement `--all` cli params as separate PR since it is independent and much less changes)

PR #28, opened 2021-01-10
Enhanced "unschedule" to remove all possible jobs

jkellerer:
This change ensures that schedules that are no longer defined in the config will still attempted to be removed from the os scheduler. It addresses one part of #27 to remove all schedules for one profile.

Notes:
* ~~Currently this is a first draft without any testing as basis for discussing the general direction.~~
* The implementation tries to be as generic as possible to support all os schedulers.
* There is no dependency on OS schedulers to list registered jobs and as a consequence, renamed profiles may still leave scheduled jobs behind. A fix for this may be added in a separate PR.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -309,7 +309,7 @@ func randomKey(output io.Writer, c *config.Config, flags commandLineFlags, args 
 
 // createSchedule accepts one argument from the commandline: --no-start
 func createSchedule(_ io.Writer, c *config.Config, flags commandLineFlags, args []string) error {
-	scheduler, profile, schedules, err := getScheduleJobs(c, flags)
+	scheduler, profile, schedules, err := mustGetScheduleJobs(c, flags)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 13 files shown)


creativeprojects:
> * The files `schedule_unix` and `schedule_windows` have the CRLF issue which increases the diff if whitespace is not ignored. Let me know when this is fixed and I'll rebase.

Argl. Sorry, I've fixed them, you can rebase 😊

jkellerer:
Rebase is done

creativeprojects:
Thanks.

So it doesn't need a schedule configuration to be able to unschedule it: only a profile name and that's all fine in this case.

Now, thinking a bit further, for `unschedule --all`, we would miss it if we changed the named of a profile.
I reckon it's not something very common but I do remember changing profile names in the past when moving backup between servers.
So we need to be able to handle that case as well. Worst case scenario we could keep a state file.

jkellerer:
Good point. I could imagine for os schedulers that support it, they could return a list of profile names which are then added in addition to known profile names for the `--all` option. That way we might still use the `systemctl list` command to get this list, but if for any reason the list cannot be generated it still works on all names that are known by the configuration.

creativeprojects:
For launchd it's a matter of reading a couple of directories and check the file names from the pattern
For Task Scheduler on Windows it's a matter of loading a list of tasks and check their names as well.

Definitely doable 👍🏻

jkellerer:
Sounds like a plan 👍 . So I'll finish this PR with unit tests unless you have some other concerns?

creativeprojects:
> Sounds like a plan 👍 . So I'll finish this PR with unit tests unless you have some other concerns?

I think that should work, no other concern I can think of

creativeprojects:
By the way @jkellerer , the schedule package was getting a bit messy so I'm doing some refactoring what will easily fix this issue.

There will be two objects:
- a Schedule object with methods on a profile
- a Job object like before

All of these accessible through interfaces as well. That's making things more extensible 👍🏻

jkellerer:
@creativeprojects , thanks a lot for the note. Haven't had time yet to continue but I'll wait for your commit first so avoid any double effort.

jkellerer:
Unit tests have been added and I've also verified it manually on MacOS & Linux. Removed WIP flag.

creativeprojects:
Cool 👍🏻 
I'll try it out some time this week
Thanks

creativeprojects:
Yep, looks good to me, thanks again 😄

jkellerer:
Thanks as well!

PR #32, opened 2021-02-07
Added fail env variable ERROR_COMMANDLINE

jkellerer:
This adds the commandline of the last failed command to the environment (as "ERROR_COMMANDLINE") to have more details on which of the calls failed (even without debug logging).


creativeprojects:
Agree, that could be handy. Nice one 👍🏻

PR #33, opened 2021-02-14
Added resticprofile flags --no-lock & --lock-wait

jkellerer:
This draft PR adds 2 resticprofile flags (and schedule & global configs) to control locking behaviour (profile locks & restic remote locks)

@creativeprojects, let me know whether this is acceptable. If so I'll test and document it. 
Currently I'm not using the built-in locking as I'm missing these 2 features.

Added flags:
* `--no-lock`: Ignores any defined locks. 
   Motivation: Not all tasks really need a lock, e.g. I usually run `resiticprofile --name $PROFILE_NAME unlock` as pre-run to ensure stale locks are removed. With profile locks this only works when locks can be ignored.

* `--lock-wait`: Wait up to the specified duration to acquire a lock before failing.
  Motivation: When triggering operations that should use a lock, one can't always avoid concurrency (e.g. in schedules that may take longer than expected). However it should be configurable whether concurrency will immediately cause a failure or only when some time passed and it didn't resolve.

Added schedule config:
* `schedule-lock-mode` (default | fail | ignore): Toggles wether to set `--lock-wait` (default) or `--no-lock` (ignore) flags.
* `schedule-lock-wait`: Toggles whether `default` sets `--lock-wait` to a duration.

Added global config:
* `restic-lock-retry-after`: Sets a delay for retrying restic commands that fail to acquire a repository lock. 0 disables handling remote lock failures.
* `restic-stale-lock-age`: Sets a lock age for detecting stale locks. Stale locks are not waited upon but `restic unlock` is tried once (when `force-inactive-lock` is also set in the profile). 0 disables stale lock handling.

Status
- [x] Implemented drafts
- [x] Test config changes
- [x] Test profile lock handling
- [x] Test restic lock handling

### DIFF
--- a/shell/command.go
+++ b/shell/command.go
@@ -65,7 +65,8 @@ func (c *Command) Run() (Summary, string, error) {
 	var err error
 	var stdout, stderr io.ReadCloser
 
-	summary := Summary{}
+	analyzer := NewOutputAnalyser()
+	summary := Summary{OutputAnalysis: analyzer}
 
 	command, args, err := c.getShellCommand()
 	if err != nil {
@@ -122,7 +123,10 @@ func (c *Command) Run() (Summary, string, error) {
 			return summary, "", err
 		}
 	}
+
+	// handle command errors
 	errors := &bytes.Buffer{}
+
 	// send error output to buffer & stderr
 	if stderr != nil {
 		stderrOutput := c.Stderr
@@ -130,14 +134,20 @@ func (c *Command) Run() (Summary, string, error) {
 			stderrOutput = os.Stderr
 		}
 		err = c.ScanStderr(stderr, stderrOutput, errors)
-		if err != nil {
-			return summary, errors.String(), err
+
+		if e := cmd.Wait(); err == nil {
+			err = e
 		}
+	} else {
+		err = cmd.Wait()
 	}
 
-	err = cmd.Wait()
+	// finish summary
 	summary.Duration = time.Since(start)
-	return summary, errors.String(), err
+	errorText := errors.String()
+	analyzer.AnalyzeStringLines(errorText)
+
+	return summary, errorText, err
 }
 
 // getShellCommand transforms the command line and arguments to be launched via a shell (sh or cmd.exe)
--- a/shell/command.go
+++ b/shell/command.go
@@ -10,6 +10,8 @@ import (
 	"runtime"
 	"strings"
 	"time"
+
+	"github.com/creativeprojects/clog"
 )
 
 const (
@@ -133,9 +135,13 @@ func (c *Command) Run() (Summary, string, error) {
 		if stderrOutput == nil {
 			stderrOutput = os.Stderr
 		}
-		err = c.ScanStderr(stderr, stderrOutput, errors)
 
-		if e := cmd.Wait(); err == nil {
+		err = c.ScanStderr(stderr, stderrOutput, errors)
+		if err != nil {
+			clog.Errorf("failed reading stderr from command: %s ; Cause: %s", command, err.Error())
+		}
+
+		if e := cmd.Wait(); e != nil {
 			err = e
 		}
 	} else {
--- a/shell/command.go
+++ b/shell/command.go
@@ -65,8 +65,8 @@ func (c *Command) Run() (Summary, string, error) {
 	var err error
 	var stdout, stderr io.ReadCloser
 
-	analyzer := NewOutputAnalyser()
-	summary := Summary{OutputAnalysis: analyzer}
+	analyser := NewOutputAnalyser()
+	summary := Summary{OutputAnalysis: analyser}
 
 	command, args, err := c.getShellCommand()
 	if err != nil {
@@ -145,7 +145,7 @@ func (c *Command) Run() (Summary, string, error) {
 	// finish summary
 	summary.Duration = time.Since(start)
 	errorText := errors.String()
-	analyzer.AnalyzeStringLines(errorText)
+	analyser.AnalyseStringLines(errorText)
 
 	return summary, errorText, err
 }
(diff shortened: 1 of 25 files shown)


creativeprojects:
Yeah I don't particularly need it the way I use restic. But if you need it on your workflow, please go ahead 😃

creativeprojects:
Actually I had this issue recently, although this PR wouldn't have solved it:

I have 2 central backup machines where all my servers push new file every night. One of the server had a big upgrade and pushed a lot more than usual, and then all the other night backup failed.

I think the "wait for lock" option could also be checking the restic lock regularly and wait until it's free to go. If possible. Maybe.

jkellerer:
Yes I have this as a separate pre-run script to check the restic lock (and wait for it and/or remove stale locks). 
That's also on my list for an additional PR. 

Sorry for the delay, try to finish this PR next week and will then also check how to integrate restic lock handling.

creativeprojects:
> Yes I have this as a separate pre-run script to check the restic lock (and wait for it and/or remove stale locks).
> That's also on my list for an additional PR.

Excellent 👍🏻 

> Sorry for the delay, try to finish this PR next week and will then also check how to integrate restic lock handling.

Don't worry, I've also been busy on other things lately 😄

jkellerer:
This PR is updated to contain missing ideas like schedule config and restic lock handling (stale and lock wait). I still need to test and review the changes myself.

jkellerer:
Unit tests have been written and code was cleaned-up.
Testing it now on 2 similar servers with overlapping schedules where my previous shell pre-script for lock-wait was removed.

jkellerer:
Had several profiles scheduled hourly on 2 servers pointing to the same repo, resulting in several 100 backups over the past days. There was lots of log output like this:
```
2021/04/20 00:23:12 lock wait (remaining 22m2s / elapsed 22m58s): rest:https://**:**@backup-server:8080/shared-repo/ locked by PID 8475 on app-serv-02 by root (UID 0, GID 0)
2021/04/20 00:15:17 lock wait (remaining 29m57s / elapsed 15m3s): /tmp/resticprofile.shared-repo.lock locked by root on Tuesday, 20-Apr-21 00:00:03 CEST from app-serv-01
```
... but not a single failure, all scheduled jobs succeeded.
Profile configuration was adjusted to contain `schedule-lock-wait: 45m` & `lock: /tmp/file`.

@creativeprojects : I'm removing the WIP flag as I think it is ready for review now.

creativeprojects:
Excellent!

I'll study it in the next few days 👍🏻 

Thanks

shell/command.go
```diff
@@ -122,22 +123,31 @@ func (c *Command) Run() (Summary, string, error) {
...
 		stderrOutput := c.Stderr
 		if stderrOutput == nil {
 			stderrOutput = os.Stderr
 		}
 		err = c.ScanStderr(stderr, stderrOutput, errors)
-		if err != nil {
-			return summary, errors.String(), err
+
+		if e := cmd.Wait(); err == nil {
+			err = e
```
jkellerer, line 139: Might be more correct to change the logic and always set err when e != nil. 
When cmd.Wait() returns an error it should be returned to get the error code of the process.
Had done it this way to preserve the original logic for error reporting (but in the original code, cmd.Wait was not called in this case)

creativeprojects: I see what you mean, if there's an error reading from the pipe it's going to end the process and probably killing restic before the end.

That is a very unlikely event but I definitely overlooked it. You're right it's better to wait until the end of the command before reporting the error 👍🏻

wrapper.go
```diff
@@ -48,15 +50,31 @@ func (c *commandError) Stderr() string {
 	return c.stderr
 }
 
+func (c *commandError) ExitCode() (int, error) {
+	if exitError, ok := asExitError(c.err); ok {
+		return exitError.ExitCode(), nil
+	} else {
+		return 0, errors.New("not and exit error. Exit code not available")
```
creativeprojects, line 57: typo: "not an exit error"
(I know I'm a typo master myself!)

shell/command.go
jkellerer: Ok. Will update it.

wrapper.go
jkellerer: Think can be shortened to "exit code not available"

creativeprojects:
> ```
> 2021/04/20 00:23:12 lock wait (remaining 22m2s / elapsed 22m58s): rest:https://**:**@backup-server:8080/shared-repo/ locked by PID 8475 on app-serv-02 by root (UID 0, GID 0)
> 2021/04/20 00:15:17 lock wait (remaining 29m57s / elapsed 15m3s): /tmp/resticprofile.shared-repo.lock locked by root on Tuesday, 20-Apr-21 00:00:03 CEST from app-serv-01
> ```

The message I'm using for the local lock is completely different than the restic one. I never noticed that before and it does look stupid 😆 (I might want to change that at some point.)

Anyway, I tried a few very close backup overnight and it worked perfectly fine, so I'm merging the PR.
I'll keep it running for a few days and release a new version later.

Thank you very much, it works nicely and it's going to be very useful 👍🏻

renard:
Coming late after the party.

Wouldn't it be useful to have options such as `--no-lock` available in the configuration file?
For example I have a `ro-repository` which is on a read-only remote ZFS volume. Locks could not be created.

I prefer having  global `lock-mode` similar to `schedule-lock-mode` instead of having to remember to add the `--no-lock` flag (which is still useful, don't get me wrong). Same with `--lock-wait`.

In the same way wouldn't it be nice to have an `extra-flags` key for commands to allow users to pass extra arguments to restic?

renard:
Ah yes that is it. We should be careful about that: the `no-lock` key MUST be `true`.

Looks like other variants are not taken in account ie. you cannot use `yes` instead of `true`.

Some implementations allow `yes`/`no`, `on`/`off`, `0`/`1` but they are not part of yaml standard:
* https://yaml.org/spec/1.2.2/#10212-boolean
* https://yaml.org/type/bool.html

This is still unclear to me. However specifying `true` instead is fine. Thanks again for the clarification.

PR #48, opened 2021-04-24
Summary from plain output when not run in terminal

jkellerer:
Small enhancement: Allow to use _restic_ plain output to fill summary (and `status-file`) when `resticprofile` output is redirected (= not sent to terminal), since in this case there is no disadvantage of intercepting stdout.

The alternative of using `extended-status: true` is even more limited as it cuts all normal output, which can be problematic when one wants to inspect the logs later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1391,9 +1391,17 @@ Here's an example of a generated file, where you can see that the last `check` f
 }
 ```
 
-## Extended status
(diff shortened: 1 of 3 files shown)


creativeprojects:
This is a good idea 😉 
I left the code commented out as I was considering reusing restic display code to send the status back to the terminal, or even just do a similar kind of output.

Thanks 👍🏻

jkellerer:
Thanks as well

PR #49, opened 2021-04-24
Updated contrib script systemd/send-error.sh

jkellerer:
Applied latest changes from ERROR env and added the profile to the mail


creativeprojects:
Thanks for the update

PR #50, opened 2021-05-16
Added contrib/grafana dashboard

jkellerer:
Here is a small contribution of my Grafana dashboard for backup monitoring (using syslog shipping as datasource).


creativeprojects:
I tried it and it's pretty cool 👍🏻 

Thanks for all the detailed explanation

PR #55, opened 2021-07-17
add Hostname pre-defined variable to template resolver

lionelnicolas:
One use case for that could be :

```
global:
  default-command: snapshots
  initialize: false
  scheduler: crond

default:
  repository: s3:http://my.local.minio/{{ .Hostname }}-{{ .Profile.Name }}
  password-file: {{ .ConfigDir }}/{{ .Hostname }}.key

(...)

home-src:
  inherit: default

  backup:
    source:
    - {{ .Env.HOME }}/src

```

Then, running `resticprofile --name home-src` would use different buckets (and keys) if running the same profile on different machines.


creativeprojects:
Looks good to me 👍🏻 Thanks

PR #58, opened 2021-08-04
Feature: Hide confidential values in output

jkellerer:
This PR wraps possibly confidential config values so that they have a public (non-confidential) and an internal value. 
The internal value is passed to `restic` while the public value is used in logging and profile output (`resticprofile show`)

**Motivation**:
* When using a rest server, basic auth credentials are usually part of the repo-url.
* In most other cases keys, tokens or passwords are part of the environment.
* Secrets may appear in the log output, when printing the profile and in error messages sent with a script like ["resticprofile-send-error.sh"](https://github.com/creativeprojects/resticprofile/tree/master/contrib/systemd)

 @creativeprojects  Let me know what you think. It probably has to wait for #57 as there will be conflicts on the same files.

**When the PR is applied, the output changes**:
```
➜ ./resticprofile --name=p3 show     
2021/08/04 19:46:42 using configuration file: profiles.yaml
p3:
    repository:     rest:https://user:***@remotehost
    password-file:  test-repo.key
    initialize:     true
    status-file:    test-repo-status.json

    env:
        my_secret_key:  ***

➜ ./resticprofile --name=p3 --dry-run                       
2021/08/04 19:46:52 using configuration file: profiles.yaml
2021/08/04 19:46:52 profile 'p3': initializing repository (if not existing)
2021/08/04 19:46:52 dry-run: /usr/local/bin/restic init --password-file test-repo.key --repo rest:https://user:***@remotehost
2021/08/04 19:46:52 profile 'p3': starting 'snapshots'
2021/08/04 19:46:52 dry-run: /usr/local/bin/restic snapshots --password-file test-repo.key --repo rest:https://user:***@remotehost
2021/08/04 19:46:52 profile 'p3': finished 'snapshots'
```

### DIFF
--- a/config/confidential.go
+++ b/config/confidential.go
@@ -1,0 +1,132 @@
+package config
+
+import (
+	"reflect"
+	"regexp"
+)
+
+const ConfidentialReplacement = "***"
+
+// ConfidentialValue is a string value with a public and a confidential representation
+type ConfidentialValue struct {
+	public, confidential string
+}
+
+func (c ConfidentialValue) Value() string {
+	return c.confidential
+}
+
+func (c ConfidentialValue) String() string {
+	return c.public
+}
+
+func (c *ConfidentialValue) IsConfidential() bool {
+	return c.public != c.confidential
+}
+
+// hideValue hides the entire value in the public representation
+func (c *ConfidentialValue) hideValue() {
+	if c.IsConfidential() {
+		return
+	}
+	c.public = ConfidentialReplacement
+}
+
+// hideSubmatches hides all occurrences of submatches in the public representation
+func (c *ConfidentialValue) hideSubmatches(pattern *regexp.Regexp) {
+	if c.IsConfidential() {
+		return
+	}
+
+	if matches := pattern.FindStringSubmatchIndex(c.confidential); matches != nil && len(matches) > 2 {
+		c.public = c.confidential
+
+		for i := len(matches) - 2; i > 1; i -= 2 {
+			start := matches[i]
+			end := matches[i+1]
+
+			c.public = c.public[0:start] + ConfidentialReplacement + c.public[end:]
+		}
+	}
+}
+
+func NewConfidentialValue(value string) ConfidentialValue {
+	return ConfidentialValue{public: value, confidential: value}
+}
+
+// confidentialValueDecoder implements parsing config parsing for ConfidentialValue
+func confidentialValueDecoder() func(from reflect.Type, to reflect.Type, data interface{}) (interface{}, error) {
+	confidentialValueType := reflect.TypeOf(ConfidentialValue{})
+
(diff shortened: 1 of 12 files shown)


creativeprojects:
Cool, I was planning on doing it at some point 😄 

I will indeed merge #57 first, in the next day or two 👍🏻

jkellerer:
Ready for review :)

jkellerer:
Ready for review (after adding additional tests for stringify and fmt.Sprintf)

config/confidential.go
```diff
@@ -0,0 +1,132 @@
...
+			value.hideValue()
+			profile.Environment[name] = value
+		} else if urlEnvKeys.MatchString(name) {
+			value.hideSubmatches(urlConfidentialPart)
+			profile.Environment[name] = value
+		}
+	}
+}
+
+// GetNonConfidentialValues returns a new list with confidential values being replaced with their public representation
```
creativeprojects, line 108: I was working on something on the side, which I pushed to the `shell-escape` branch if you want to have a look.

Basically, instead of having a `map[string][]string` to build a list of arguments, I'm now using a collection (shell.Args) of arguments (shell.Arg) which contain the information on how the argument should be escaped to the shell.

I'm wondering if it's something that could be merged together (information about confidentiality as well as how to escape it to the command line).

**But** that wouldn't satisfy the *separation of concern* principle very well, would it 🤔 

Although I suppose I could still use a `ConfidentialValue` instead of a `string` inside `shell.Arg`. In which case that would simplify the way we construct the real and public command line (and we would have no need for this method <- which was the point of the comment 😄 )

creativeprojects (commented):
It looks good to me, it all works on my profiles 👍🏻

config/confidential.go
jkellerer: First of all: I'm glad that you're working on proper arg escaping 👍. This bit me once but could be solved with `files-from` (and some pre script that creates a list).

To the comment, I see the point. `GetNonConfidentialValues` is a bit of a hacky way to convert an argument list with possible secret values to a public argument list. It is not the best solution and when arguments are escaped it might even break. 

The method could be converted to operate on shell.Args (as input and output) in which case it would have access to the raw values and could map them to public values.

The better approach would probably be to generate a public and an actual list from the sources but it requires changes to the methods that generate shell.Args.

I also think that concerns must be kept separate, otherwise it gets very dirty. So shell.Args should not need to know about public and confidential values. Effectively we have a public and confidential shell.Args instance, either by updating the existing method or by generating the 2 from the config.

creativeprojects: The implementation works fine on the current codebase, so I'm going to merge it.
Now I'll try to find a nice way to implement proper argument escaping 😄 

Thanks 👍🏻

Issue #56, opened 2021-07-23
[Bug] Arguments not properly being escaped


Syphdias:
I noticed problems when using `dump` with file paths that have spaces in them.
This is not just a problem of `dump` but of all commands and also not just limited to to spaces but everything that needs to be quoted. Symbols like `'` and `"` come to mind which can be part of a file name.

This can be observed via `--dry-run` but appears that this option builds it's output independently from what is actually executed. This can also be used to execute arbitrary commands. I don't know if this would be exploitable in any way – probably not – but is definitely not how it should behave.
```
❯ resticprofile --config /tmp/profiles.toml -n local --dry-run dump 1231245125 "/path/with space; echo foo"
2021/07/23 01:56:23 profile 'local': starting 'dump'
2021/07/23 01:56:23 dry-run: /usr/bin/restic dump --password-file /home/restic/restic.local.password --repo /restic 1231245125 /path/with space; echo foo
2021/07/23 01:56:23 profile 'local': finished 'dump'
❯ resticprofile --config /tmp/profiles.toml -n local dump 1231245125 "/path/with space; echo foo"
2021/07/23 01:56:27 profile 'local': starting 'dump'
Fatal: no file and no snapshot ID specified
foo
2021/07/23 01:56:27 profile 'local': finished 'dump'
```

I think I could track the bug down to [`shell/command.go` line 184](https://github.com/creativeprojects/resticprofile/blob/f58116ba38473d1e702c55423f03f223626ea829/shell/command.go#L182):
```golang
    return shell, []string{"-c", strings.Join(flatCommand, " ")}, nil
```

I don't quite get the need for why it needs to be called with `sh -c` and not just directly via the first argument in `flatCommand`. Compatibility? Path variable? Something like this should do:
```golang
    return flatCommand[0], flatCommand[1:], nil
```
Also the dry run should use this code instead of building it's own interpretation of the arguments.

creativeprojects:
There's a simple reason why I had to use `sh -c "restic ..."`:
The expectation of resticprofile is **not** to change the way we would interact with restic. Imagine you would normally run this kind of command:
```
$ restic backup /data*
```

The shell will expand the path before executing the command, and restic will receive (as an example) `restic backup /data0 /data1 /data2`

To be able to have this path expansion, I need to call restic through the shell. And it is not particularly convenient.

The other option would be to do the path expansion myself (and maybe some other things the shell would do to the command line). But I'm not entirely sure of the complexity of it.

[...]

PR #60, opened 2021-08-06
Shell escape

creativeprojects:
The current implementation of passing (funny) parameters to restic is broken, as explained in #56 

Let's take this config as an example:

```yaml
---
default:
  password-file: key
  repository: "rest:http://user:password@localhost:8000/path"

simple:
  inherit: default
  backup:
    exclude: "/**/.git"
    source: /source

spaces:
  inherit: default
  backup:
    exclude: "My Documents"
    source: /source dir

glob:
  inherit: default
  backup:
    exclude:
      - "examples/integration*"
    source:
      - "examples/integration*"

quote:
  inherit: default
  backup:
    source:
      - "/Côte d'Ivoire"
```

For each profile, here are the differences between the current implementation and the fix:

## simple

This is a simple configuration with no funny character (no glob, no space, just plain). **This is what most people will have as a configuration.**

`resticprofile -n spaces backup --option`

Current and new implementations are sending the same arguments, there's no change. Here's the list of arguments restic will receive:

`backup`
`--exclude`
`/**/.git`
`--password-file`
`key`
`--repo`
`rest:http://user:password@localhost:8000/path`
`--option`
`/source`

## spaces

Current implementation fails if you add some parameters containing spaces on the command line:

`resticprofile -n spaces backup "some path"`

Current | Fix
--- | ---
`backup` | `backup`
`--exclude` | `--exclude`
`My Documents` | `My Documents`
`--password-file` | `--password-file`
`key` | `key`
`--repo` | `--repo`
`rest:http://user:password@localhost:8000/path` | `rest:http://user:password@localhost:8000/path`
`some` | `some path`
`path` |
`/source dir` | `/source dir`

## glob

`resticprofile -n glob backup examples/integration*`

Let's imagine we have 2 files in `examples/` named `integration_test.toml` and `integration_test.yaml`

Current implementation is working fine in this case, no change.

`backup`
`--exclude`
`examples/integration*`
`--password-file`
`key`
`--repo`
`rest:http://user:password@localhost:8000/path`
`examples/integration_test.toml`
`examples/integration_test.yaml`
`examples/integration_test.toml`
`examples/integration_test.yaml`

## quote

`resticprofile -n quote backup "/path/with space; echo foo"`

Adding a single quote in a name actually breaks the current implementation with the shell returning an error 2 (there's no closing quote)

Current | Fix
--- | ---
`backup` | `backup`
`--password-file` | `--password-file`
`key` | `key`
`--repo` | `--repo`
`rest:http://user:password@localhost:8000/path` | `rest:http://user:password@localhost:8000/path`
`/path/with` | `/path/with space; echo foo`
`space;` |
`echo` |
`foo` |
`/Côte d'Ivoire` | `/Côte d'Ivoire`

**Unless anybody objects to this new implementation, I will merge the fix soon-ish** 👍🏻


jkellerer:
Regarding the `source` list defined in a profile: I think this should always be escaped and as a consequence this requires to resolve glob expressions internally using `filepath.Glob()` instead of relying on shell expansion.

Consider the following example:
```yaml
----
profglob:
  repository: rest:https://user:pass@remotehost
  password-file: test-repo.key

  retention:
    after-backup: true
    keep-last: 3

  backup:
    source: path*
```

This results in the following command lines when running `./resticprofile -n profglob --dry-run backup`:
````
/usr/local/bin/restic backup --password-file test-repo.key --repo rest:https://user:***@remotehost path*
/usr/local/bin/restic forget --keep-last 3 --password-file test-repo.key --path /test/path* --repo rest:https://user:***
````

For `backup` this works fine, but in `forget` it breaks, since all paths must be written as "--path path1 --path path2 --path pathX ...".

Note: It may also make sense (but this is out of scope for this PR) to introduce `path` to `retention` which can be set to `false` (not add a path filter), `true` (current behaviour) and list of paths overriding `source`. In general it makes sense that `retention` filters by paths but for glob expressions it may not since paths are not stable (dynamic) and snapshots may not match.

File: shell/arg.go [hunk 1]
```diff
@@ -0,0 +1,110 @@
...
+	doubleQuotePattern = regexp.MustCompile(`[^\\][|&;<>()$'" \t\r\n*?[]`)
+)
+
+type ArgType int
+
+const ArgTypeCount = 4
+
+const (
+	ArgConfigEscape       ArgType = iota // escape each special character but don't add quotes
+	ArgConfigNoGlobQuote                 // use double quotes around argument when needed
```
jkellerer, line 22: The name NoGlob (and "no-glob") made me think at first that glob expressions are forbidden, but in fact it means that glob expressions are passed to restic (as required for include/exclude patterns).

I know this is just internal, but would it make sense to use a name like "preserve-glob" or "keep-glob"?

creativeprojects: You're right, that helps with code readability 😛

creativeprojects:
> 
> For `backup` this works fine, but in `forget` it breaks, since all paths must be written as "--path path1 --path path2 --path pathX ...".
> 

You see, it's funny because I knew I was forgetting something but couldn't figure out what, hence why I was leaving the PR opened 🤔

PR #63, opened 2021-08-13
Resolve glob expressions in backup sources

jkellerer:
Resolves glob expressions in `source` internally instead of relying on shell expansion. 
This fixes problems with `retention` and `path` where shell expansion doesn't work.

~~Marked as draft until #60 is merged, also misses a test at the moment.~~

### DIFF
--- a/config/path.go
+++ b/config/path.go
@@ -31,6 +31,59 @@ func fixPaths(sources []string, callbacks ...pathFix) []string {
 	return fixed
 }
 
+func expandEnv(value string) string {
+	if strings.Contains(value, "$") || strings.Contains(value, "%") {
+		value = os.ExpandEnv(value)
+	}
+	return value
+}
+
+func isUserHomePath(value string) bool {
+	// Only "~" and "~/path" but not "~somefile"
+	return strings.HasPrefix(value, "~") &&
+		(len(value) < 2 || strings.ContainsAny(value[1:2], `/\`))
+}
+
+func expandUserHome(value string) string {
+	if isUserHomePath(value) {
+		if home, err := os.UserHomeDir(); err == nil {
+			value = home + value[1:]
+		} else {
+			clog.Warningf("cannot resolve user home dir for path %s : %v", value, err)
+		}
+	}
+	return value
+}
+
+func absolutePrefix(prefix string) pathFix {
+	return func(value string) string {
+		if value == "" ||
+			filepath.IsAbs(value) ||
+			isUserHomePath(value) ||
+			strings.HasPrefix(value, "$") ||
+			strings.HasPrefix(value, "%") {
+			return value
+		}
+		return filepath.Join(prefix, value)
+	}
+}
+
+func absolutePath(value string) string {
+	if value == "" ||
+		filepath.IsAbs(value) ||
+		isUserHomePath(value) ||
+		strings.HasPrefix(value, "$") ||
+		strings.HasPrefix(value, "%") {
+		return value
+	}
+	if absolute, err := filepath.Abs(value); err == nil {
+		return absolute
+	}
+	// looks like we can't get an absolute version...
+	clog.Errorf("cannot determine absolute path for '%s'", value)
+	return value
+}
+
 // resolveGlob evaluates glob expressions in a slice of paths and returns a resolved slice
 func resolveGlob(sources []string) []string {
 	resolved := make([]string, 0, len(sources))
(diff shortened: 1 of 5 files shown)


jkellerer:
Hi @creativeprojects , the PR changes the behaviour with glob expressions in `Source` and this caused `TestFromConfigFileToCommandLine/examples\integration_test.*/glob2/backup ` to fail on Windows due to escaping happening on path separators (\). 

While updating the test I found that probably the way how paths are escaped on Windows is not completely correct. The backslash "\" is not an escape character in Windows (for `cmd.exe` it is [^] and for `powershell` it is [`] as far as I know).

The issue could be worked around for the moment by ensuring that `func resolveGlob()` converts paths to use slashes. This would fix the test but path escaping on Windows would still not be right, I think.

creativeprojects:
> Hi @creativeprojects , the PR changes the behaviour with glob expressions in `Source` and this caused `TestFromConfigFileToCommandLine/examples\integration_test.*/glob2/backup ` to fail on Windows due to escaping happening on path separators [\].
> 
> While updating the test I found that probably the way how paths are escaped on Windows is not completely correct. The backslash [\] is not an escape character in Windows (for `cmd.exe` it is [^] and for `powershell` it is [`] as far as I know).
> 
> The issue could be worked around for the moment by ensuring that `func resolveGlob()` converts paths to use slashes. This would fix the test but path escaping on Windows would still not be right, I think.

hmmm I would think there's no need to escape any character in Windows, as the parameters can be sent one by one (and not all as one parameter to `sh`).
I'm no longer using Windows actually, but I have a VM for testing. I might have missed something, I need to check that carefully again.

jkellerer:
After checking test again, I found that paths are already compared against c-style escaped expected paths (e.g. `"examples\\key"` contained in a literal string):
https://github.com/creativeprojects/resticprofile/blob/3d82291ddae1c8c6d02df160e907100b032d97a5/integration_test.go#L58

The fact that paths use double backslash on Windows in the test don't seem to be related to `shell.Args` as I've seen that escaping in `Arg.String` is properly skipped as you also mentioned. Have no Windows platform either to fully debug this. 

However since the behaviour is unrelated to the changes of this PR (and also not related to `shell.Args`), I think it can be ignored. For that reason I've adjusted the test to c-style escape expected glob paths for Windows to pass the test (paths are now consistent as other expected paths use the same c-style escaping).

creativeprojects:
It does work on Windows, although I'm not sure why it's expecting a double `\` I'm a bit puzzled as well to be honest.

On the other OSes, it doesn't work for a backup path like `~/Doc*`: _filepath.Glob_ cannot resolve it.
We could use something like _filesearch.ShellExpand_ but we can't make a difference between a path with a space and distincts values, unless we can find a shell trick for that.

jkellerer:
Thanks a lot for the hint with home-path. While shell expand would be an option, I think it is better to handle paths internally and not pass them to the shell (as there may be several side effects of doing so). There still is `files-from` to pass-in a file-list which can be generated by a `run-before` script if resolving through the shell is required as input.

I've added `expandUserHome` and updated tests. (Note: Also powershell supports "~", therefore I wouldn't skip it on Windows)

creativeprojects:
Thanks! it looks good 👍🏻 
Now I'm sorry but I'm going to be annoying: should we support the lesser known `~user/` format? (home directory of a specific user)

I'm also thinking, if we do expand all the arguments this way, there's no need for shell escaping anymore since we can call restic directly without going via `sh /c` (and send all arguments one bye one like for Windows)

jkellerer:
Not being biggest fan of `ShellExpand` in this case I think it's the only way to support the special case `~username` properly.  Implementation & test is updated.

creativeprojects:
If you prefer, we should be able to use user.Lookup to get a *User object with Homedir property. I had a quick look and it reads the information from `/etc/passwd` so it should be right.
I might also replace the ShellExpand in which case, I think the `~` is the only funny character not managed by filepath.Glob

(there's a last thing, `~+` and `~-` but surely nobody is going to use that, that's too much)

jkellerer:
Thanks again for the hint, haven't had the `user` package on my mind. PR is updated.

config/path.go
```diff
@@ -66,3 +101,20 @@ func absolutePath(value string) string {
 	clog.Errorf("cannot determine absolute path for '%s'", value)
 	return value
 }
+
+// resolveGlob evaluates glob expressions in a slice of paths and returns a resolved slice
+func resolveGlob(sources []string) []string {
+	resolved := make([]string, 0, len(sources))
+	for _, source := range sources {
+		if strings.ContainsAny(source, "?*") {
```
creativeprojects, line 109: I think matching characters should be "?\*[" since you can use expressions like [aA]*

creativeprojects (changes requested):
Thanks for all the work!

I think we just need to add the "[" to the list of characters in ContainsAny and it all looks good to me.

config/path.go
jkellerer: Thanks a lot for the review. Not using this very often with glob so I overlooked it. Tests are updated as well.

creativeprojects (approved):
Good work! thanks 😄

PR #64, opened 2021-08-13
Allow overriding "default-command" in profile

jkellerer:
Add "default-command" from `global` also to the profile sections.
This allows to create profiles (and group of such profiles) for special cases, e.g. like `check` or `prune`:
```yaml
groups:
  prune-all: ["prune-repo1", ...]

prune-repo1:
  inherit: "repo1"
  default-command: "prune"
  prune:
    schedule: "daily"
...
```

The example above can be run with `resticprofile -n prune-all`

### DIFF
--- a/config/profile.go
+++ b/config/profile.go
@@ -35,6 +35,7 @@ type Profile struct {
 	PrometheusSaveToFile string                       `mapstructure:"prometheus-save-to-file"`
 	PrometheusPush       string                       `mapstructure:"prometheus-push"`
 	PrometheusLabels     map[string]string            `mapstructure:"prometheus-labels"`
+	DefaultCommand       string                       `mapstructure:"default-command"`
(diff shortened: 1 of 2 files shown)


creativeprojects:
This is one way of doing it, while keeping the `groups` section the way it is, which is fine.

Also, I have been thinking about the groups configuration for a while, and I clearly overlooked it when I did it.
At some point I would like to change it so we can configure options inside a group, the most obvious would be to be able to schedule a group.

In which case, default command would fit nicely in a group configuration.

Now I don't mind adding a default command per profile though

jkellerer:
Having "default-command" in a group is also a good thing. Priority wise it's just a question what comes first when we have `gobal`, `group` and `profile`. When it is added to groups as well it is probably best to prioritise `group`, otherwise profiles with a "default-command" cannot be added to groups that are built for scheduling a particular purpose.

PR #65, opened 2021-08-13
Supporting config `includes` (e.g. `profiles.d`)

jkellerer:
This PR adds support for splitting the profile configuration into multiple config files that can be included from the main "profiles" config file (glob expressions are supported for `conf.d` or `profiles.d` style configs).

The implementation allows to mix any supported config file format and merges things together. HCL format is special in that it doesn't support mixing with other formats.

Example `profiles.toml`:
````toml
##
# Loading config overrides from "conf.d" and profiles from "profiles.d":
includes = [
  'conf.d/*.conf',
  'profiles.d/*.toml',
  'profiles.d/*.yaml',
]

##
# Defaults
[default]
repository = "/backup"
password-file = "{{.ConfigDir}}/backup.password"
````

### DIFF
--- a/README.md
+++ b/README.md
@@ -625,6 +625,24 @@ resticprofile will search for your configuration file in these folders:
 - c:\resticprofile\
 - %USERPROFILE%\
 
+# Includes
+
+The configuration may be split into multiple files by adding `includes = "glob-pattern"` to the main configuration file. 
+E.g. the following `profiles.conf` loads configurations from `conf.d` and `profiles.d`:
+```toml
+# Includes
+includes = ["conf.d/*.conf", "profiles.d/*.yaml", "profiles.d/*.toml"]
+
+# Defaults
+[global]
+initialize = true
+...
+```
+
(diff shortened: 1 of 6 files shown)


creativeprojects:
Sorry I was a bit busy lately,

I quite like this idea 👍🏻 

One thing I couldn't find any information via a quick search: do we know if filepath.Glob always return the files in the same order?

creativeprojects:
So that wasn't too difficult after all: https://github.com/golang/go/issues/17153

I think we should order the results, just in case 😄

jkellerer:
Btw. while I added this primarily for deb/rpm/apk packages, it can also be helpful when there is a REST API some day for management. Such an API could be a frontend to a folder structure that builds the config for `resticprofile` running in a server role as part of a client & server model. E.g. folders like "conf.d", repositories.d", "generic-profiles.d" could be the backend storage for "config", "repository" and "profile" REST endpoints.

PR #67, opened 2021-08-15
Optional: Allow disabling path in retention with 'false'

jkellerer:
This change is really optional and more a question whether it makes the built-in behaviour of path copy clearer when it can be toggled explicitly.

Remaining questions for me are:
* Would it make sense to apply such a logic to all commands that support `path` as parameter? (Other command would then default to `false` to not change their behaviour)
* Would it make sense to treat `tags` in a similar way?

### DIFF
--- a/README.md
+++ b/README.md
@@ -477,8 +477,9 @@ keep-within = "3h"
 keep-tag = [ "forever" ]
 compact = false
 prune = false
-# if path is NOT specified, it will be copied from the 'backup' source
-# path = []
+# path can be a boolean ('true' meaning to copy source paths from 'backup') 
+# or a path or list of paths to use instead. Default is `true` if not specified.
+#path = []
 # the tags are NOT copied from the 'backup' command
 tag = [ "test", "dev" ]
 # host can be a boolean ('true' meaning current hostname) or a string to specify a different hostname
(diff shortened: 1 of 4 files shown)


creativeprojects:
I don't see any particular issue with it, but it means we just treat the `retention` section like `forget`; the only difference is that we don't offer the `before-backup` and `after-backup` on the `forget` config section.

jkellerer:
> but it means we just treat the retention section like forget
True. Hmm.. maybe for the moment we leave it the way it is.

Does this PR make sense?
.. and does it make sense to support `tags` copy in the `retention` section (as it already is done for paths)?

creativeprojects:
Yes, I think this PR makes sense if we want to copy the `tags` during a `forget`, then that leaves the user a choice of `paths` or `tags` (or both).
Annoyingly the default for `path` is `true` whereas the default for `tags` would be `false` (not to change the existing behaviour)

jkellerer:
Yes the fact that tags and paths have different defaults is a bit annoying, however host defaults to "true" and combined with paths it is a relative exact filter for retention, so the effect of this selection of defaults is not that serious (as long as os.Hostname is correct and backup paths do not overlap) 

Will extend the PR to also copy tags

creativeprojects (approved):

PR #70, opened 2021-08-18
Support "run-finally" in backup-section & profile

jkellerer:
This PR adds a new post command `run-finally` that is meant to run always regardless of what happened before. 

The main purpose is to have a hook for resource cleanup, e.g. consider the following profile:

```yaml
default:
  backup:
    source: "/opt/*/_backup"
    run-before:
      - btrfs subvolume snapshot -r /opt/apps/ /opt/apps/_backup
      - btrfs subvolume snapshot -r /opt/services/ /opt/services/_backup
    run-finally:
      - btrfs subvolume delete /opt/apps/_backup
      - btrfs subvolume delete /opt/services/_backup
```

The current available hooks `run-after` and `run-after-fail` would require to add the cleanup calls twice and may leave snapshots behind if one of the calls has a non-zero exit code. Also `run-after-fail` is not available for backup (and doesn't have to be).

Note: I thought about calling it `run-after-all` and `run-always` but used `run-finally` as it describes the behaviour best. Commands in this list are run after any other commands and are are always started, even if there is a panic somewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -645,23 +645,28 @@ documents:
   run-before: "echo == run-before profile $PROFILE_NAME command $PROFILE_COMMAND"
   run-after: "echo == run-after profile $PROFILE_NAME command $PROFILE_COMMAND"
   run-after-fail: "echo == Error in profile $PROFILE_NAME command $PROFILE_COMMAND: $ERROR"
+  run-finally: "echo == run-finally $PROFILE_NAME command $PROFILE_COMMAND"
   backup:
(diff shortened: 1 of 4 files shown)


creativeprojects:
It's a good idea, I like the name `run-finally`.

At some point I'll rework the configuration a bit, make a `version: '2'` like docker compose files, and I'll rename the two others `run-on-success` and `run-on-failure` or something better like that. But that's for later.

creativeprojects (approved):
It looks good to me :+1

PR #84, opened 2021-11-10
Fix args are multiplied when commands are retried

jkellerer:
Discovered that when commands are retried (e.g. `backup` with lock failure) the command line contains repeated arguments:
````
"/usr/bin/restic" "backup" ... "--repo" "rest:https://user:×××@repo/" \
"--tag" "etc" "--tag" "home" "/etc" "/home" "/root" "/etc" "/home" \
"/root" "/etc" "/home" "/root" "/etc" "/home" "/root" "/etc" "/home" ...
````

This PR fixes it :)


creativeprojects:
Certainly is a funny bug! thanks for that 😄

Issue #52, opened 2021-06-26
[Discussion] Regarding systemd units (includes feature requests/ideas)


Syphdias:
Hi, I started writing my own restic wrapping for running restic via systemd. After that I found `resticprofile`. I am currently in the process of replacing my systemd units with resticprofile ones but ran into some questions. I don't mean this as "hey you should do things differently and my way is the right way" but I do have some suggestions and hope you don't mind.

For my use case: I use 2 restic repos ("s3" and "local") on my machine and backup `/` and `$HOME` in different intervals. I also do regular `restic forget`s and `restic prune`s for both repositories.

Legend: I use "job" for "backup", "check", "forget" and "prune". Not sure what it is actually called I am trying to avoid "command".

# timers and units
By default a timer called `foo@bar.timer` or `baz.timer` will activate `foo@bar.service` or `baz.service` respectively. You define it explicitly with `Unit=` which is not necessary unless you activate a service which has another name. I haven't see this anywhere with `resticprofile`, yet.
This also works with instances of templates (see below).

# activating/scheduling
Currently there is no `--dry-run` option for `resticprofile schedule` and `resticprofile unschedule` as far as I can tell. And it seems to be impossible to only activate certain jobs (e.g. backup) but not others (check) without adjusting config.
Also I'd like to be able to create the unit files, but _not_ activate them right away (symlinking to `timers.target.wants` directory) but only creating the service and timer units. (More of a feature request than a discussion)

# systemd templating
I noticed you use the systemd template syntax but you don't utilize it. What I mean by that is that you use `template@instance` (for timer or service) but you create a file for each of them.
An example how this could be done with a template (only including important parts, this not an actual unit):
```
# /etc/systemd/system/resticprofile-check@.service
[Service]
ExecStart=/usr/local/bin/resticprofile --no-prio --no-ansi --config /etc/resticprofile/profiles.toml --name %i check
```
This has the benefit of uncoupling the scheduling and activating from `resticprofile`. You could still use the `resticprofile schedule` but you would only need one template for each job (backup, check, etc.). This also means you can activate jobs for profiles that have no schedule without additional config needed.

## timer templating
The same concept works for timers but not as well because there are three things to take into account: profile, job and schedule.
[...]

PR #89, opened 2022-02-06
Add run flag to to be able to have profile and job name as one argument

Syphdias:
This implements the idea from #52 to make it easy to write systemd unit templates that directly correspond to a defined resticprofile job.

* `--run` takes `<profile>.<job>`
* if there is another `.` is will be interpreted as part of the job name
* overrides `--name`
* `--` for extra `restic` parameters

I am not sure how to write tests for this. Could you give me some pointers how to do so and what should be tested for?

### DIFF
--- a/README.md
+++ b/README.md
@@ -943,7 +943,7 @@ There are not many options on the command line, most of the options are in the c
 * **[-f | --format] configuration_format**: Specify the configuration file format: `toml`, `yaml`, `json` or `hcl`
 * **[-n | --name] profile_name**: Profile section to use from the configuration file.
   You can also use `[profile_name].[command]` syntax instead, this will only work if `-n` is not set.
-  Bot using `-n [profile_name] [command]` and `[profile_name].[command]` are supported.
+  Using `-n [profile_name] [command]` or `[profile_name].[command]` both select profile and command and are technically equivalent.
 * **[--dry-run]**: Doesn't run the restic command but display the command line instead
 * **[-q | --quiet]**: Force resticprofile and restic to be quiet (override any configuration from the profile)
 * **[-v | --verbose]**: Force resticprofile and restic to be verbose (override any configuration from the profile)
--- a/README.md
+++ b/README.md
@@ -871,11 +871,19 @@ Backup root & src profiles (using _full-backup_ group shown earlier)
 ```
 $ resticprofile --name "full-backup" backup
 ```
+or
+```
+$ resticprofile full-backup.backup
+```
 
 Assuming the _stdin_ profile from the configuration file shown before, the command to send a mysqldump to the backup is as simple as:
 
 ```
 $ mysqldump --all-databases | resticprofile --name stdin backup
+```
+or
+```
+$ mysqldump --all-databases | resticprofile stdin.backup
 ```
 
 Mount the default profile (_default_) in /mnt/restic:
@@ -890,8 +898,8 @@ Display quick help
 $ resticprofile --help
 
 Usage of resticprofile:
-	resticprofile [resticprofile flags] [restic command] [restic flags]
-	resticprofile [resticprofile flags] [resticprofile command] [command specific flags]
+	resticprofile [resticprofile flags] [profile name.][restic command] [restic flags]
+	resticprofile [resticprofile flags] [profile name.][resticprofile command] [command specific flags]
 
 resticprofile flags:
   -c, --config string        configuration file (default "profiles")
@@ -933,7 +941,9 @@ There are not many options on the command line, most of the options are in the c
 * **[-h]**: Display quick help
 * **[-c | --config] configuration_file**: Specify a configuration file other than the default
 * **[-f | --format] configuration_format**: Specify the configuration file format: `toml`, `yaml`, `json` or `hcl`
-* **[-n | --name] profile_name**: Profile section to use from the configuration file
+* **[-n | --name] profile_name**: Profile section to use from the configuration file.
+  You can also use `[profile_name].[command]` syntax instead, this will only work if `-n` is not set.
+  Bot using `-n [profile_name] [command]` and `[profile_name].[command]` are supported.
 * **[--dry-run]**: Doesn't run the restic command but display the command line instead
 * **[-q | --quiet]**: Force resticprofile and restic to be quiet (override any configuration from the profile)
 * **[-v | --verbose]**: Force resticprofile and restic to be verbose (override any configuration from the profile)
--- a/flags.go
+++ b/flags.go
@@ -40,8 +40,8 @@ func loadFlags(args []string) (*pflag.FlagSet, commandLineFlags, error) {
 
 	flagset.Usage = func() {
 		fmt.Println("\nUsage of resticprofile:")
-		fmt.Println("\tresticprofile [resticprofile flags] [restic command] [restic flags]")
-		fmt.Println("\tresticprofile [resticprofile flags] [resticprofile command] [command specific flags]")
+		fmt.Println("\tresticprofile [resticprofile flags] [profile name.][restic command] [restic flags]")
+		fmt.Println("\tresticprofile [resticprofile flags] [profile name.][resticprofile command] [command specific flags]")
 		fmt.Println("\nresticprofile flags:")
 		flagset.PrintDefaults()
 		fmt.Println("\nresticprofile own commands:")
@@ -94,13 +94,13 @@ func loadFlags(args []string) (*pflag.FlagSet, commandLineFlags, error) {
 	flags.resticArgs = flagset.Args()
 
 	// if there are no further arguments, no further parsing is needed
-	if (len(flags.resticArgs) == 0) {
+	if len(flags.resticArgs) == 0 {
 		return flagset, flags, err
 	}
 
 	// parse first postitional argument as <profile>.<command> if the profile was not set via name
 	nameFlag := flagset.Lookup("name")
-	if ((nameFlag == nil || !nameFlag.Changed) && strings.Contains(flags.resticArgs[0] , ".")) {
+	if (nameFlag == nil || !nameFlag.Changed) && strings.Contains(flags.resticArgs[0] , ".") {
 		// split first argument at `.`
 		profileAndCommand := strings.Split(flags.resticArgs[0], ".")
 		// last element will be used as restic command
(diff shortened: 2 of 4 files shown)


creativeprojects:
Hi!

I tried to push a new commit to your branch but apparently I'm not allowed => So I added some code in the master branch (that you can rebase).

If you couldn't find how to run unit tests, it's because, you ... couldn't. At least not without a small change.

I changed the signature of the `loadFlags` method like so:

```go
func loadFlags(args []string) (*pflag.FlagSet, commandLineFlags, error) {
}
```

which means it no longer takes the arguments directly from the command line (`os.Args()`), but as a parameter.

You can see, I added some simple tests in the new file `flags_test.go` and you can follow the same model 😄 

Thanks 👍🏻

Syphdias:
> which means it no longer takes the arguments directly from the command line (`os.Args()`), but as a parameter.

Not sure, if that will affect my code or not, yet.

> You can see, I added some simple tests in the new file `flags_test.go` and you can follow the same model :smile:

That's perfect! Writing the first test actually made me realize – without even running it – that I described what I wanted to do with a second dot correctly but implemented it the wrong way round.

I will shortly rebase onto `master` and the write some tests. But somehow `go test` fails for me on `master` without any modification :thinking:

Syphdias:
Nevermind that failed test. I forgot to stash changes in `flags_test.go`

creativeprojects:
> That's perfect! Writing the first test actually made me realize – without even running it – that I described what I wanted to do with a second dot correctly but implemented it the wrong way round.

This is precisely why we write tests 😄 it looks simple but somehow we still manage to fail a corner case 👍🏻

Syphdias:
> This is precisely why we write tests smile it looks simple but somehow we still manage to fail a corner case 👍🏻

That's also why I then wrote a test for the parsing I wanted and then changed the code :wink: 

> I tried to push a new commit to your branch but apparently I'm not allowed => So I added some code in the master branch (that you can rebase).

Btw, I did tick "maintainer can edit".

So do you have any feedback for what could be done better?

flags.go
```diff
@@ -91,5 +94,18 @@ func loadFlags(args []string) (*pflag.FlagSet, commandLineFlags, error) {
 	// remaining flags
 	flags.resticArgs = flagset.Args()
 
+	if flags.run != "" {
+		// abort if argument not of the right format
+		if strings.Count(flags.run, ".") < 1 {
+			fmt.Fprintf(os.Stderr, "Please provide profile name and command separated by \".\", e.g. default.check\n")
+			os.Exit(1)
```
creativeprojects, line 101: That would be better to return an error here, like
```go
return flagset, errors.New("please provide profile name and command separated by \".\", e.g. default.check")
```

1. it's not good practise to stop the execution of the whole program when an error arises inside a function
2. with this you can add a test to verify that a wrong argument returns an error (so we test both the _happy path_ and the _sad path_)

flags_test.go
```diff
@@ -36,3 +36,12 @@ func TestBackupProfileName(t *testing.T) {
...
 }
+
+func TestBackupProfileNameAsRun(t *testing.T) {
+	_, flags, err := loadFlags([]string{"-r", "prof.ile1.backup", "--verbose", "--", "-v"})
+	require.NoError(t, err)
+	assert.False(t, flags.help)
+	assert.Equal(t, flags.name, "prof.ile1")
+	assert.True(t, flags.verbose)
+	assert.Equal(t, flags.resticArgs, []string{"backup", "-v"})
+}
```
creativeprojects, line 47: I would probably add another test with these arguments `[]string{"-r", "prof.ile1.backup", "--", "-v"}` and make sure the verbose flag is **not** set, so we can be sure it's not picked up from the `-v` we send directly to restic

and maybe another test to make sure a simple `[]string{"-r", "profile1.backup"}` is well understood

(you can never run enough tests 😆 )

creativeprojects (changes requested):

flags.go
Syphdias: Raising error and improved parsing to check for more _sad paths_.

flags_test.go
Syphdias: I did, though the Quality Gate seems to think they are too similar to write twice :grimacing:

jkellerer:
@Syphdias, @creativeprojects, this is a great addition, nevertheless I've got the following thoughts on this PR:

* So far commands don't contain a dot. Only groups and profile names might contain it. Wouldn't it be better to always take the last "." as delimiter and allow dots in profile or group name?
* Would it make sense to allow the following input `profile.` (aka _profile.default-command_), `[.]command` (aka _default-profile.command_) in addition to `profile-or-group.command`
* Further would it make sense to apply the logic to the command string to support ".", instead of adding a `--run`flag. E.g.:

  | Commandline                            | Profile (or Group) | Command         |
  |----------------------------------------|--------------------|-----------------|
  | `resticprofile profile.command`        | profile            | command         |
  | `resticprofile --name profile command` | profile            | command         |
  | `resticprofile profile.`               | profile            | default-command |
  | `resticprofile --name profile`         | profile            | default-command |
  | `resticprofile [.]command`             | default-profile    | command         |

What do you think?

Syphdias:
> @Syphdias if you remote the `- flags.go` line in `codecov.yml` it will pick up your tests. I had it disabled so it wasn't complaining about `flags.go` not being tested (there was no code in there before, only flag definition)

Done

> Yeah you're right, it's more simple this way, the profile can be added in front of the command and that doesn't break the way it works now 👍🏻

I like it, I can probably get to rewriting it tomorrow.

Syphdias:
Should `resticprofile .` be invalid or default profile + default command?

creativeprojects:
Which is the same as `resticprofile` with no argument at all.

It kinds of look weird, but I don't mind either way.

It's not needed but it could help with scripts that omit variables, something like that, where both variables are empty:
```
resticprofile ${profile}.${command}
```

Syphdias:
> ```
> resticprofile ${profile}.${command}
> ```

If this is bash, I would recommend `set -u` to disallow using undefined variables. Unfortunately bash has lots of Gotchas like that.

I had one thing I couldn't solve and marked it with `FIXME`.
Usually using `-n someProfile` results in the first argument to be taken as is; `resticprofile -n someProfile foo.bar`: use "someProfile" as profile and `restic foo.bar` as command.
But I could not figure out how to detect if `-n` was used to I check if it is the default value. This results in no override for using `-n default`; `resticprofile -n default foo.bar`: "foo" as profile and `restic bar` as command.

Any ideas?

jkellerer:
> 
> Any ideas?

The following should do this:
```go
if nameFlag := flags.Lookup("name"); nameFlag != nil && nameFlag.Changed {
	// User has set --name or -n
}
```

Syphdias:
> ```go
> flag := flagset.Lookup("name"); flag != nil && flag.Changed
> ```

If I understand this correctly, `flag` can never be `nil`since it has a default value. In case this should ever change I still included it in the condition.

jkellerer:
> If I understand this correctly, `flag` can never be `nil`since it has a default value. In case this should ever change I still included it in the condition.

Right it can never be `nil` since it is declared as flag. However it is better to check for `nil` where it is valid to return it as code may change or bugs may produce `nil` where it isn't expected.

Syphdias:
I think I am mostly happy with this. I renamed some of the tests to better reflect what they test.
Let me know what you think and what needs to get changed.

Syphdias:
Rebased for clean history

jkellerer:
Look good to me. The only little uncommon thing for `go` is the use of outer parentheses in `if` statements. They are usually not added.

jkellerer:
Guess what we might need is a short note in README.md and a small hint in CLI help, e.g.:

```go
fmt.Println("\tresticprofile [resticprofile flags] [profile name.][restic command] [restic flags]")
fmt.Println("\tresticprofile [resticprofile flags] [profile name.][resticprofile command] [command specific flags]")
```

creativeprojects:
> Looks good to me. Besides this very well tested

I fully agree 😄

> Guess what we might need is a short note in README.md and a small hint in CLI help

That's right, maybe we can put a little introduction of the new syntax in the README, with a confirmation that the existing syntax is still supported.

jkellerer:
> that the existing syntax is still supported

.. would say the new syntax is an alternative to the "existing syntax". Both have there use cases. E.g. in shell completion it can get quite noisy with larger profiles.yaml and adding `--name` shrinks number of possible candidates.

Syphdias:
> Guess what we might need is a short note in README.md and a small hint in CLI help, e.g.:

Absolutely agree.

> That's right, maybe we can put a little introduction of the new syntax in the README, with a confirmation that the existing syntax is still supported.

Will do. Note that if `--name` is set, the behavior is completely identical to before. There might be arguments to adjust this behavior in the future, e.g. `--name foo .backup` could still just be the same as `--name foo backup`.

> ```go
> fmt.Println("\tresticprofile [resticprofile flags] [profile name.][restic command] [restic flags]")
> fmt.Println("\tresticprofile [resticprofile flags] [profile name.][resticprofile command] [command specific flags]")
> ```

I think technically it should be this since `.something` will be interpreted as defaultprofile + `something` command.
```go
fmt.Println("\tresticprofile [resticprofile flags] [[profile name].][restic command] [restic flags]")
fmt.Println("\tresticprofile [resticprofile flags] [[profile name].][resticprofile command] [command specific flags]")
```
But I think it does not add anything useful. I cannot thing of a use-case for `.check` over just `check`.

README.md
```diff
@@ -933,7 +941,9 @@ There are not many options on the command line, most of the options are in the c
 * **[-h]**: Display quick help
 * **[-c | --config] configuration_file**: Specify a configuration file other than the default
 * **[-f | --format] configuration_format**: Specify the configuration file format: `toml`, `yaml`, `json` or `hcl`
-* **[-n | --name] profile_name**: Profile section to use from the configuration file
+* **[-n | --name] profile_name**: Profile section to use from the configuration file.
+  You can also use `[profile_name].[command]` syntax instead, this will only work if `-n` is not set.
+  Bot using `-n [profile_name] [command]` and `[profile_name].[command]` are supported.
```
jkellerer, line 946: Small typo in this line. how about:

Using `-n [profile_name] [command]` or `[profile_name].[command]` both select profile and command and are technically equivalent.

jkellerer: Thanks!

jkellerer:
@creativeprojects , think it can be merged now

creativeprojects:
@Syphdias @jkellerer thank you both for the PR 👍🏻

Issue #86, opened 2022-01-23
Bash completions


djmattyg007:
It would be great to have an official bash completion script.

creativeprojects:
Hi,
Yes I thought about it, it's on my TODO list.

But I haven't had a chance to look at how to generate completion files.
And since I'm not using cobra so I can't rely on it to generate them for me 😞

jkellerer:
Played a bit with bash completion today just for fun resulting in a working proof of concept (#90).

@creativeprojects what do you think, does this fit your plan of going forward or did you want to add `cobra` instead?

creativeprojects:
> Played a bit with bash completion today just for fun resulting in a working proof of concept (#90).
> 
> what do you think, does this fit your plan of going forward or did you want to add `cobra` instead?

So far, it does look good to me 😛 

I was looking at cobra, and it's going to mess up the separation of flags between resticprofile and restic and I wasn't too keen on searching some kind of hack to make it work.

And we don't really need cobra anyway, I find it a bit annoying to use.

So yeah, generating a shell script is the way to go, thank you very much 👍🏻

creativeprojects:
PR is merged, completion will be available in version 0.17.0 👍🏻

PR #90, opened 2022-02-06
Shell completion

jkellerer:
This PR provides shell completion (addressing #86) using a custom `complete` command operating on the data structures in resticprofile rather than auto-generating a complete script (like restic does) which would require more refactoring on resticprofile's internals.

Since `resticprofile` is a wrapper around `restic`, all unknown input is forwarded to shell completion of `restic` and both results are then merged.

**Setup**

```shell
resticprofile completion-script --bash > /etc/bash_completion.d/resticprofile
chmod +x /etc/bash_completion.d/resticprofile
source /etc/bash_completion.d/resticprofile
```

or

```shell
eval "$(resticprofile completion-script --bash)"
```

**Example**

```shell
bash-5.1$ resticprofile 
backup             diff               init               profiles           schedule           tag
cache              dump               key                prune              self-update        unlock
cat                find               list               random-key         show               unschedule
check              forget             ls                 rebuild-index      snapshots          version
copy               generate           migrate            recover            stats              
create-completion  help               mount              restore            status             

bash-5.1$ resticprofile -
--config       --help         --name         --no-prio      --theme        --wait         -h             -q
--dry-run      --lock-wait    --no-ansi      --quiet        --trace        -c             -l             -v
--format       --log          --no-lock      --self-update  --verbose      -f             -n             -w

bash-5.1$ resticprofile --config pr
priority       profiles.yaml  progress       prom     

bash-5.1$ resticprofile --config profiles.yaml --no-
--no-ansi  --no-lock  --no-prio  

bash-5.1$ resticprofile --config profiles.yaml --no-lock --theme 
dark   light  none   

bash-5.1$ resticprofile --config profiles.yaml --no-lock --theme dark schedule --
--all       --no-start  

bash-5.1$ resticprofile mount -
--allow-other             --limit-download=         --password-file=          --tls-client-cert
--cacert                  --limit-upload            --path                    --tls-client-cert=
--cacert=                 --limit-upload=           --path=                   --verbose
--cache-dir               --no-cache                --quiet                   -H
--cache-dir=              --no-default-permissions  --repo                    -o
--cleanup-cache           --no-lock                 --repo=                   -p
--host                    --option                  --repository-file         -q
--host=                   --option=                 --repository-file=        -r
--json                    --owner-root              --snapshot-template       -v
--key-hint                --password-command        --snapshot-template=      
--key-hint=               --password-command=       --tag                     
--limit-download          --password-file           --tag=                    
```

# ToDo

- [x] Support bash
- [x] Support zsh (basic only)
- [x] Unit test `complete` function and fix bugs
- [x] Document

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ $ make install
 
 Shell commandline completions are provided for `bash` and `zsh`. 
 
-To load completions in the shell (user) profile, use:
+To load completions in shell or profile, use:
 
 ```shell
 # bash
@@ -267,8 +267,6 @@ To install them permanently:
 $ resticprofile completion-script --bash > /etc/bash_completion.d/resticprofile
 $ chmod +x /etc/bash_completion.d/resticprofile
 ```
-
-Note: Check path, might be `/etc/...` or `/usr/local/etc/...`
 
 
 # Upgrade
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ For the rest of the documentation, I'll be showing examples using different form
   * [Manual installation (Windows)](#manual-installation-windows)
   * [Ansible](#ansible)
   * [Installation from source](#installation-from-source)
+  * [Shell completion](#shell-completion)
 * [Upgrade](#upgrade)
 * [Using docker image](#using-docker-image)
   * [Container host name](#container-host-name)
@@ -244,6 +245,31 @@ To install the binary in your user path:
 ```
 $ make install
 ```
+
+
+## Shell completion
+
+Shell commandline completions are provided for `bash` and `zsh`. 
+
+To load completions in the shell (user) profile, use:
+
+```shell
+# bash
+eval "$(resticprofile completion-script --bash)"
+
+# zsh
+eval "$(resticprofile completion-script --zsh)"
+```
+
+To install them permanently:
+
+```
+$ resticprofile completion-script --bash > /etc/bash_completion.d/resticprofile
+$ chmod +x /etc/bash_completion.d/resticprofile
+```
+
+Note: Check path, might be `/etc/...` or `/usr/local/etc/...`
+
 
 # Upgrade
 
--- a/commands.go
+++ b/commands.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"crypto/rand"
+	_ "embed"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -28,12 +29,19 @@ type ownCommand struct {
 	description       string
 	action            func(io.Writer, *config.Config, commandLineFlags, []string) error
 	needConfiguration bool              // true if the action needs a configuration file loaded
-	hide              bool              // don't display the command in the help
+	hide              bool              // don't display the command in help and completion
+	hideInCompletion  bool              // don't display the command in completion
 	flags             map[string]string // own command flags should be simple enough to be handled manually for now
 }
 
-var (
-	ownCommands = []ownCommand{
+var ownCommands []ownCommand
+
+func init() {
+	ownCommands = getOwnCommands()
+}
+
+func getOwnCommands() []ownCommand {
+	return []ownCommand{
 		{
 			name:              "version",
 			description:       "display version (run in verbose mode for detailed information)",
@@ -93,7 +101,26 @@ var (
 			hide:              false,
 			flags:             map[string]string{"--all": "display the status of all scheduled jobs of all profiles"},
 		},
+		{
+			name:              "completion-script",
+			description:       "generate a shell completion script (use --bash or --zsh to choose format)",
+			action:            completionScriptCommand,
+			needConfiguration: false,
+			hide:              false,
+			hideInCompletion:  true,
+			flags: map[string]string{
+				"--zsh":  "create zsh completion script",
+				"--bash": "create bash completion script",
+			},
+		},
 		// hidden commands
+		{
+			name:              "complete",
+			description:       "create commandline completion results based on given args",
+			action:            completeCommand,
+			needConfiguration: false,
+			hide:              true,
+		},
 		{
 			name:              "elevation",
 			description:       "test windows elevated mode",
@@ -116,7 +143,7 @@ var (
 			hide:              true,
 		},
 	}
-)
+}
 
 func displayOwnCommands(output io.Writer) {
 	commandsWriter := tabwriter.NewWriter(output, 0, 0, 3, ' ', 0)
@@ -239,6 +266,31 @@ func panicCommand(_ io.Writer, _ *config.Config, _ commandLineFlags, _ []string)
 
 func testCommand(_ io.Writer, _ *config.Config, _ commandLineFlags, _ []string) error {
 	clog.Info("Nothing to test")
+	return nil
+}
+
+func completeCommand(output io.Writer, _ *config.Config, _ commandLineFlags, args []string) error {
+	completions := (&Completer{}).Complete(args)
+	if len(completions) > 0 {
+		for _, completion := range completions {
+			fmt.Fprintln(output, completion)
+		}
+	}
+	return nil
+}
+
+//go:embed contrib/completion/bash-completion.sh
+var bashCompletionScript string
+
+//go:embed contrib/completion/zsh-completion.sh
+var zshCompletionScript string
+
+func completionScriptCommand(output io.Writer, _ *config.Config, _ commandLineFlags, args []string) error {
+	if containsString(args, "--zsh") {
+		fmt.Fprintln(output, zshCompletionScript)
+	} else {
+		fmt.Fprintln(output, bashCompletionScript)
+	}
 	return nil
 }
 
--- a/commands.go
+++ b/commands.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"runtime"
 	"runtime/debug"
 	"sort"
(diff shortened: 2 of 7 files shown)


creativeprojects:
Thank you for working on this, much appreciated 👍🏻 

For the other PR I changed the `loadFlags` function  so we can run unit test on it, it might be helpful for you as well actually 😄

jkellerer:
Cool, then I'll go ahead and create tests and missing bits. Regarding `loadFlags`: Yes I've seen it and can revert my customisation (only did them to prevent exit on error).

djmattyg007:
I noticed that the PR has a todo item for zsh. I don’t mean to dismiss zsh as a perfectly good shell, but if it’s going to take a lot of extra work you might want to consider leaving that for a follow-up PR so that bash users can get some value out of this sooner.

jkellerer:
Yes, zsh will get basic (or no) support in this PR and a followup PR will fix zsh. Just delayed writing of tests until #89 is sorted (which it is now).

jkellerer:
Windows tests fail on parsing `examples/dev.yaml`:

```plain
185:         source:
186:         - "C:\Users\runneradmin/go/src"
187:         
...
cannot parse yaml configuration: While parsing config: yaml: line 186: did not find expected hexdecimal number
```

Caused by:

```yaml
         source:
         - "{{ .Env.HOME }}/go/src"
```

May need to be evaluated at some day but has nothing todo with this PR. Will use a different conf file to fix tests on Windows.

jkellerer:
Ready for review now. Bash is fully supported, zsh has basic support.

Please note that the PR assumes the changes from #89 and as such profile prefixes are offered as completions.

jkellerer:
Updated profile prefix completion after reviewing #89 and finding it works slightly different (and in a better way) than assumed.

creativeprojects:
I've finally installed bash-completion on my linux laptop. I guess I'm old fashion unix after all 😆 

There's still a few things I want to try and I'll complete the review 👍🏻

jkellerer:
Completion should work on any modern bash, only restic delegation requires bash completion to be installed

creativeprojects:
It works perfectly fine with bash (tried on a Debian machine) 👍🏻 

I tried with zsh on my Mac but it fails. I understand it's using the bash compatibility, but still can't get it to work.

Have you managed to make it work for zsh?
It's ok if it's not ready for zsh though, but then in that case we should disable the --zsh flag.

Here's what I get:

```
% eval "$(./resticprofile completion-script --zsh)"
zsh: no such file or directory: 2022/02/25
 %
```

and when I try _manually_ (with `bashcompinit` loaded) I get this instead:

```
% . ./resticprofile completion-script --bash
[2] 23935
./resticprofile:1: no such file or directory: Z^D^@^@^@^@Z^C^@^@^@^@^@^@^F
[2]  + exit 1      EZ�� < ZZ <
./resticprofile:1: bad pattern: EZ^@^E^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@__typelink^@^@^@^@^@^@__TEXT^@^@^@^@^@^@^@^@^@^@\M-@\M-v\M-^@^D^@^@^@^@0O^@^@^@^@^@^@\M-@\M-v\M-^@^@^E^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@__itablink^@^@^@^@^@^@__TEXT^@^@^@^@^@^@^@^@^@^@^@F\M-^A^D^@^@^@^@\M-x^U^@^@^@^@^@^@^@F\M-^A^@^E^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@__gosymtab^@^@^@^@^@^@__TEXT^@^@^@^@^@^@^@^@^@^@\M-x[\M-^A^D^@^@^@^@^@^@^@^@^@^@^@^@\M-x[\M-^A^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@__gopclntab^@^@^@^@^@__TEXT^@^@^@^@^@^@^@^@^@^@^@\M-^A^D^@^@^@^@\M- \M-b
./resticprofile:3: parse error near `)'
 %
```

jkellerer:
Hmm.. in my case it looks like this with zsh (on Mac as well):
````
resticprofile on  ft-shell-completion via 🐹 v1.17.6
➜ eval "$(./resticprofile completion-script --zsh)"

resticprofile on  ft-shell-completion via 🐹 v1.17.6 
➜ resticprofile --format json
hcl   json  toml  yaml
````

**Comment by jkellerer** [comment]
Maybe some more error handling is required.

**Comment by jkellerer** [comment]
Will check tomorrow if I can fix this.

**Comment by creativeprojects** [comment]
Oh I see what you mean! silly me you're right; when I tried on Debian I did install the new resticprofile in the path. But here I run it directly from the branch in the source code. My bad 👎🏻 

So, well it shouldn't happen really; it's a very corner case

**Comment by jkellerer** [comment]
.. and the latest release will start logging "2022/02/25 ..." when asked for a completion script :)

Btw. this is only an issue for zsh, bash always works since it properly detect how to call itself.

**Comment by jkellerer** [comment]
Btw. Before you merge this, I'd like to do a last cleanup on the 2 shell scripts. I found one issue with the bash script that is also just a corner case but I'd like to fix it.

**Comment by creativeprojects** [comment]
No problem 👍🏻 

It works on zsh now 😉   Except it looks like it can't find the profile names:

```
% resticprofile --config examples/dev.yaml --name
```

tab shows nothing even though the profiles can be loaded

```
% resticprofile --config examples/dev.yaml profiles

Profiles available (name, sections, description):
  default:    (copy, env)                               Contains default parameters like repository and password file
  documents:  (backup, snapshots)
  dropbox:    (backup)
  escape:     (backup, retention)
  home:       (backup)
  nobackup:   (copy)
  prom:       (backup)
  root:       (backup, forget, retention)
  self:       (backup, check, copy, forget, retention)
  space:      (n/a)                                     Repository contains space
  src:        (backup, retention, snapshots)
  stdin:      (backup, snapshots)
  system:     (backup, forget)

Groups available (name, profiles, description):
  full-backup:  [root, src]

%
```

If this is a limitation of the bash compatibility, we can live with that for now

**Comment by jkellerer** [comment]
Looks like it can load it only when "--config" option isn't needed, right. Hadn't found this so far. Would also agree that the limitation isn't so severe but I'll still check if I can fix it. It can only happen due to filtering of input since `resticprofile complete --config examples/dev.yaml --name` probably returns a list

**Comment by creativeprojects** [comment]
It certainly does

```
% resticprofile complete --config examples/dev.yaml --name
default
documents
dropbox
escape
full-backup
home
nobackup
prom
root
self
space
src
stdin
system
%
```

**Comment by jkellerer** [comment]
Code is rebased against master and the problem you had found should be fixed (last 2 commits).

**Comment by creativeprojects** [comment]
Thanks, I'll have a look at it. 👍🏻 

One more thing I wanted to ask, how would that work if we need to install a new version of the script?

I can see two ways to setup the completion:
- running `resticprofile completion-script` in the startup script (`.bashrc` or `.zshrc`)
- saving the generated file into a completion folder (like `/usr/share/bash-completion/completions/resticprofile`) and forget about it

In the first case we're good because we're always generating the latest version. In the second case however we might end up with a script that is no longer working well with the new flags or whatever we change with the command line.

My question is, is it possible (and how easy?) to add some kind of versioning into the script so we can warn the user if it's out of date?
I'm sorry I am no expert in completion script so I don't know how difficult it is, that's why I'm asking.

**Comment by jkellerer** [comment]
Yes that is a good point. I’m not an expert for completion scripts either but what we can do is passing a version number and evaluate this in the complete command.

If it is set, we can react on it. If it isn’t, we assume the latest version so that complete can be called manually for testing.

**Comment by creativeprojects** [comment]
That's what I thought, having a hidden flag to pass the script version to the complete command. It doesn't even need to do anything with it yet.

Also we might never need this, but by experience we'd better be safe than sorry 😆 

Thanks for you help on this 😉

**Comment by jkellerer** [comment]
Yes, I’ll probably use an env variable for this case since it needs to be evaluated outside of the Completer type to allow selecting an appropriate implementation (when it ever needs to be evaluated). 

At the moment I‘d just log a warning to stderr when missing.

**Comment by jkellerer** [comment]
Changed my mind while implementing versioning regarding what to do with the information. It is evaluated since newer completion scripts should not fail as resticprofile in the path is downgraded.

Works like this now:
```
➜ resticprofile complete __POS:3 schedule --all
2022/02/27 18:22:39 Requester "unknown" version 0 not explicitly supported
--no-start

➜ resticprofile complete bash:v1 __POS:3 schedule --all
--no-start

➜ resticprofile complete bash:v10 __POS:3 schedule --all

➜ resticprofile complete zsh:v1 __POS:3 schedule --all

```

jkellerer:
Thanks a lot for the good review so far.

creativeprojects:
Works fine 😄  Tested on both bash & zsh (Linux & Mac)

Really nice suite of tests 👍🏻 

Thank you very much for this excellent PR

djmattyg007:
FWIW

> /usr/share/bash-completion/completions/resticprofile

It's the job of the packager to install files into that location, not individual pieces of software. Nothing except the Linux package manager should be touching files inside of `/usr` (except `/usr/local`).

jkellerer:
Thanks for the note. That is the plan, resticprofile only provides the script it never writes it somewhere on its own.

PR #66, opened 2021-08-13
DEB, RPM, APK packages

jkellerer:
This PR attempts to create DEB, RPM and APK packages to simplify deployment on Linux. It is an early draft at the moment and requires more work and probably some iterations and feedback.

It depends on the following:
- [ ] Config layout for `/etc/resticprofile` (using similar concepts as other software)
- [ ] Post installation script (e.g. calls `resticprofile generate-key` if password file is missing)
- [ ] Update to `.goreleaser` to create `deb`, `rpm` & `apk`
- [ ] #65 
- [ ] #64 
- [ ] #62

PR #95, opened 2022-02-27
Fix config includes when any pattern has no match

jkellerer:
Small bugfix for config includes, is a cherry-pick from #66

### DIFF
--- a/filesearch/filesearch.go
+++ b/filesearch/filesearch.go
@@ -141,6 +141,7 @@ func FindConfigurationIncludes(configFile string, includes []string) ([]string, 
 	addFile := func(file string) {
 		file = filepath.FromSlash(filepath.Clean(file))
 		if file != configFile {
+			clog.Tracef("include: %s", file)
 			files = append(files, file)
 		}
 	}
@@ -159,8 +160,10 @@ func FindConfigurationIncludes(configFile string, includes []string) ([]string, 
 				for _, match := range matches {
 					addFile(match)
 				}
+			} else if err == nil {
+				clog.Debugf("no match: %s", include)
 			} else {
-				return nil, err
+				return nil, fmt.Errorf("%w: %q", err, include)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


filesearch/filesearch.go
```diff
@@ -159,7 +160,10 @@ func FindConfigurationIncludes(configFile string, includes []string) ([]string,
 				for _, match := range matches {
 					addFile(match)
 				}
+			} else if err == nil {
+				clog.Tracef("no match: %s", include)
 			} else {
+				clog.Errorf("no match: %s ; %v", include, err)
```
creativeprojects, line 166: the error is returned and displayed after `cannot load configuration file`. So I think here we should add the missing information to the error returned instead of displaying it. Like:
```
return nil, fmt.Errorf("%w: %q", err, include)
```

creativeprojects (changes requested):

filesearch/filesearch.go
jkellerer: Makes sense

filesearch/filesearch.go
```diff
@@ -159,7 +160,10 @@ func FindConfigurationIncludes(configFile string, includes []string) ([]string,
 				for _, match := range matches {
 					addFile(match)
 				}
+			} else if err == nil {
+				clog.Tracef("no match: %s", include)
```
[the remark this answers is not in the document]
jkellerer: Warning is too verbose. In #66 I have optional includes so it is not uncommon to have no matches. Maybe debug instead of trace?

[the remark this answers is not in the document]
creativeprojects: ok, works for me 👍🏻

filesearch/filesearch.go
jkellerer: Returns an error like "`syntax error in pattern: "/tmp/[--]"`" now.

creativeprojects: Perfect, thanks

creativeprojects:
LGTM!

Sunday fun... every time I type LGTM I think of this: https://www.youtube.com/watch?v=rR4n-0KYeKQ

PR #98, opened 2022-03-12
Feature: Take command output as stdin for restic

jkellerer:
PR introducing "stdin-command" to allow scheduling backups of streams (e.g. mysql-dump) without requiring temporary storage.

Usage:

```yaml
my-profile:
  backup:
    schedule: 'hourly'
    stdin-filename: mysql-dump.sql
    stdin-command:
      - mysqldump --order-by-primary --host=db-host --user=db-user --password=db-pw db-name
```

- [x] Needs rebase when PR #94 is merged 
      (depends on `ResolveConfiguration()`)
- [x] Testing (unit & manual tests)
- [x] Contains data race that requires fixing

### DIFF
--- a/wrapper.go
+++ b/wrapper.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"os/exec"
@@ -31,6 +32,7 @@ type resticWrapper struct {
 	moreArgs     []string
 	sigChan      chan os.Signal
 	setPID       func(pid int)
+	stdin        io.ReadCloser
 	progress     []progress.Receiver
 
 	// States
@@ -57,6 +59,7 @@ func newResticWrapper(
 		command:       command,
 		moreArgs:      moreArgs,
 		sigChan:       c,
+		stdin:         os.Stdin,
 		progress:      make([]progress.Receiver, 0),
 		startTime:     time.Unix(0, 0),
 		executionTime: 0,
@@ -280,10 +283,6 @@ func (r *resticWrapper) prepareCommand(command string, args *shell.Args, moreArg
 	rCommand.stdout = term.GetOutput()
 	rCommand.stderr = term.GetErrorOutput()
 
-	if command == constants.CommandBackup && r.profile.Backup != nil && r.profile.Backup.UseStdin {
-		clog.Debug("redirecting stdin to the backup")
-		rCommand.useStdin = true
-	}
 	return rCommand
 }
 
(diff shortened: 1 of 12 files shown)


creativeprojects:
I like it.

I was using a temp file, but that will make my config simpler 😄 

```toml

    [azure.backup]
    limit-upload = 500
    exclude = [ "azure-backup.log" ]
    exclude-file = [ "root-excludes", "excludes" ]
    exclude-caches = true
    tag = [ "mysql", "git" ]
    run-before = [
        "rm -f /tmp/mysqldump.gz",
        "mysqldump -u{{ mysql_backup_user }} -p{{ mysql_backup_password }} --opt zabbix | gzip > /tmp/mysqldump.gz"
    ]
    source = [ "/tmp/mysqldump.gz", "/git/creativeprojects" ]
    run-after = [
        "rm -f /tmp/mysqldump.gz"
    ]
    schedule = "04:18"
    schedule-permission = "system"
    schedule-log = "azure-backup.log"

```

wrapper.go
```diff
@@ -324,6 +332,16 @@ func (r *resticWrapper) runCommand(command string) error {
...
+			// Redirect a stream source to stdin of restic if configured
+			if source, err := r.prepareStreamSource(); err == nil {
+				if source != nil {
+					streamSource = source
+					rCommand.stdin = streamSource
+				}
+			} else {
+				return newCommandError(rCommand, "", fmt.Errorf("%s on profile '%s': %w", r.command, r.profile.Name, err))
+			}
 		}
```
creativeprojects, line 345: I noticed when running with --dry-run there's a race condition where the main command sometimes finishes before displaying the stdin command(s)

creativeprojects (approved):
it is a bit annoying, but I think we have to synchronise the stdin goroutine with the command runner during a `--dry-run`

wrapper.go
jkellerer: Will have a look

jkellerer: The sync leads to a better implementation as `restic` is only started after the command returned initial bytes. My attempts to fix this pass all local tests but fail in CI. Will check it later.

jkellerer: After lots of fiddling, I finally found that `/usr/bin/sh -c "child-proc ..."` on Ubuntu has issues to propagate SIGINT signals to the child-proc (Ubuntu uses "dash" for "sh"). On unix "bash" is now used if available (with fallback to "sh") and it fixes the problem.

On MacOS "sh" maps to "bash" therefore the tests had succeeded there.

wrapper_streamsource.go
```diff
@@ -0,0 +1,167 @@
...
+		reader: io.MultiReader(initialReader, pipeReader),
+
+		read: func(w *readerWrapper, bytes []byte) (n int, err error) {
+			n, err = w.reader.Read(bytes)
+
+			// Stopping restic when stream source command fails while producing content
+			if err != nil && err != io.EOF {
+				clog.Errorf("interrupting '%s' after stdin read error: %s", r.command, err)
+				if runtime.GOOS == "windows" {
+					os.Exit(1) // no signalling on Windows, use os.Exit() instead
```
creativeprojects, line 126: This is brutal. I can't think of a better way though 😞

creativeprojects (approved):
That'd be nice to find a better way to stop on Windows but I'm not even sure we can do it.
Apart from that, it's all good for me, nice job 😊

wrapper_streamsource.go
creativeprojects: Good news! a simple `return` does work on Windows: that closes stdin and stops restic gracefully.
Sorry it took me a while to test it as I found out Virtualbox is no longer working with Monterey so I had to dust off an old laptop 😄

Issue #103, opened 2022-04-18
JSON Schema


DanielRamp:
**Is there a [JSON Schema](https://json-schema.org) for restic(profile) configuration files?**

I see a JSON Schema especially useful if the [editor supports it](https://json-schema.org/implementations.html#editors) (e.g. [VS Code](https://code.visualstudio.com)), because the editing of a supported file is much more comfortable.

some examples from other projects you might know:
- JavaScript [`package.json`](https://json.schemastore.org/package.json)
- [`docker-compose.yaml`](https://raw.githubusercontent.com/compose-spec/compose-spec/master/schema/compose-spec.json)

_I couldn't find any information about an existing schema for resticprofile, so maybe there isn't one?_
_If this is not yet planned, this issue is more a feature request/enhancement idea._

creativeprojects:
No, there's no schema available yet. But I was planning to do so.

I need to look up how to do it, as I've never published any before

jkellerer:
Had a quick look on libs that can generate a schema (as a starting point):
* https://github.com/atombender/go-jsonschema - very simple implementation, could be taken as source and adjusted to work on `mapstructure` tags.
* https://github.com/invopop/jsonschema - is more complete, maintained and should also be usable on the structs when defining an `AdditionalFields` callback that maps `mapstructure` tags to `json` tags.

creativeprojects:
Oh excellent!

It means little maintenance. I like that 😉 

Maybe we could also read the cobra flags from restic repo so they get added automatically (wishful thinking at best).

jkellerer:
> Maybe we could also read the cobra flags from restic repo so they get added automatically (wishful thinking at best)

This is not so hard, I think. Cobra generates manual pages that are quite well structured and not too difficult to parse. From there we can get commands, options and even descriptions. We'd only need to decide if that happens during build time (my preference), runtime or both (when manual pages exist already). Build time means we'd need to reference restic as dependency and build it. As a side effect this enables real integration tests.

Knowing all restic commands and valid options would enable more than just json-schema, it also allows to validate the config, filter CLI args (e.g. replace the static list I added few days ago) and improve shell completion.

creativeprojects:
> Knowing all restic commands and valid options would enable more than just json-schema, it also allows to validate the config, filter CLI args (e.g. replace the static list I added few days ago) and improve shell completion.

[...]

Issue #96, opened 2022-03-04
copy cmd. not recognising password2


chymian:
version 16.01.

after manually initialization of repo 2 with the --copy-chunker-params, the following copy cmd of a repo does not work:
```
resticprofile --name=home copy 
2022/03/04 09:25:03 profile 'home': initializing repository (if not existing)
2022/03/04 09:25:03 starting command: /usr/bin/restic init --option rclone.program=/usr/bin/rclone --option rclone.args=serve\ restic\ --stdio\ --transfers\ 10\ --verbose --password-file key --repo /dpool/backup/restic/sagar/home   
2022/03/04 09:25:03 profile 'home': starting 'copy'
2022/03/04 09:25:03 starting command: /usr/bin/restic copy --option rclone.program=/usr/bin/rclone --option rclone.args=serve\ restic\ --stdio\ --transfers\ 10\ --verbose --password-file key --password-file2 key --repo /dpool/backup/restic/sagar/home --repo2 rclone:s3:/sagar/restic//home                                                            
repository 92c570cd opened successfully, password is correct
rclone: 2022/03/04 09:25:10 NOTICE: S3 bucket sagar path restic//home: Odd name received "restic/home/keys/8457008b6c6e88ad02793634a8a0c3b6afc5f9c94324e75d72a370ae0a82b1f1"
Fatal: wrong password or no key found
2022/03/04 09:25:10 starting 'run-after-fail' profile command 1/1
2022/03/04 09:25:10 copy on profile 'home': exit status 1
```

Manually resorting the parameters to the following works:
```
$ restic  --option rclone.program=/usr/bin/rclone --option rclone.args="serve restic --stdio --transfers 10 --verbose" --password-file key --repo /dpool/backup/restic/sagar/home copy --repo2 rclone:s3:/sagar/restic/home  --password-file2 key
repository 92c570cd opened successfully, password is correct
repository 79e3bbb9 opened successfully, password is correct
…
```
#### another question
is there a way to automatically let resticprofile initialize the copy target-repo with the --copy-chunkers-params flag?

creativeprojects:
> after manually initialization of repo 2 with the --copy-chunker-params, the following copy cmd of a repo does not work:

I can't replicate it with my setup: I'll try with `rclone` as it looks like it could be the position of the `options` that makes it fail.

> is there a way to automatically let resticprofile initialize the copy target-repo with the --copy-chunkers-params flag?

[...]

PR #107, opened 2022-04-24
JSON Schema

jkellerer:
**Implements**:
* #103 

**Fixes**: 
* #142 
* #133 (by sending only supported flags to restic)
* #166
* ~~#96 (order of other flags to support options)~~ (separate PR)

- [x] Dictionary of `restic` commands and options (with descriptions, defaults, type etc.)
- [x] Config info API (descriptions, defaults, restic(profile) option, etc.)
- [x] JSON schema for V2 config
- [x] JSON schema for V1 config
- [x] Test it
- [x] Add command for the functionality


jkellerer:
Just a small update on the next steps:

Will created JSON schema directly from the info API as this is easier to test and implement than trying to use the lib that I introduced before. After that commands and all tests will be implemented.

Scope of this PR will be: 
* JSON schema and command to generate it
* Command to process a template against the info API (to create a config file reference)
* Simplify profile.go further, now that info API knows the structure already

Out of scope (to be delivered later):
* Profile validation
* More intelligent processing of restic options (only pass options that a command really understands)

jkellerer:
Depends on https://github.com/creativeprojects/go-selfupdate/pull/33

jkellerer:
Added "More intelligent processing of restic options" to the scope of this PR (can be toggled in `global`)

creativeprojects:
This is getting huge 😄 
I suspect you didn't realise it would get out of control when you started it 😛 

I'm moving it to `0.21.0` as I'd like to push a new docker image with the new restic in it 👍🏻 
(I might want to find a way to push a docker image without doing a build but it doesn't look too easy with goreleaser for now)

jkellerer:
Yes true, it got a bit out of control :) .. but I'm almost there, just need to restore test coverage and do some final tests with real configurations.

Nevertheless it may still take a few more weeks before it is final, so yes moving it to 0.21 makes sense if you want to release now.

jkellerer:
PR is finally implemented. I'd like to test it manually for another week before I label it ready for review.

jkellerer:
@creativeprojects , I've left all individual commits for now, but will squash everything together before merging.

Take your time and let me know what you think.

creativeprojects:
Thanks, it looks awesome!

85 files changed... it's going to take me a while though 😆

restic/commands.go
```diff
@@ -0,0 +1,575 @@
...
+				} else {
+					tc.Options = append(tc.Options, sopt)
+				}
+			}
+
+			// record removed options
+			sourceOpts := optionsSet(sc.Options)
+			for i := range tc.Options {
+				if _, exists := sourceOpts[tc.Options[i].Name]; !exists {
+					tc.Options[i].RemovedInVersion = sourceVersion
```
creativeprojects, line 332: Should we only set `RemovedInVersion` when it's empty? Otherwise it will look like options are always removed at the latest know version 😜

jkellerer: Very good catch. This fixes version information for 2 flags. Will commit the change when I have it unit-tested.

creativeprojects:
Well I have to say I'm very pleased with this!
There was a lot to do behind the scene to load everything from the restic man pages.

Also I have to admit I haven't gone through everything, mostly the parts that could break the configuration.

I'll have a few follow ups afterwards (I don't want to delay this PR), like the reference page in the documentation which is too big. I'd like to cut it in parts, but we'll do that later.

Having said that, I found 2 things so far.

1. the profile `prom` from `dev.yaml` (where I removed the tags template) is showing an error on the prometheus labels section:
```yaml
prom:
    force-inactive-lock: true
    initialize: true
    inherit: default
    prometheus-save-to-file: "self.prom"
    prometheus-push: "http://localhost:9091/"
    prometheus-labels:
      - host: "{{ .Hostname }}"
    status-file: /Volumes/RAMDisk/status.json
    backup:
        extended-status: true
        no-error-on-warning: true
        source:
          - "{{ .CurrentDir }}"
```
Screenshot 2023-02-22 at 21 32 59

The profile runs well though.

2. The `run-before` and `run-after` requires an array:
```yaml
run-before-after:
    inherit: default

    snapshots:
        host: true
        run-before: "echo Before snapshots"
        run-after: "echo After snapshots"
```

There's no error reported in vscode, but running the profile fails:

Screenshot 2023-02-22 at 21 38 33

jkellerer:
To 1.) Yes as you wrote: `- key: value` means `[]map[string]string` but we just have a single map not an array. So the json schema should not allow it.

To 2.) Very good catch, again. The schema validation is correct but I missed to pass the correct `mapstructure.DecoderConfig` for dynamic sections like `snapshot`, therefore decoding fails. Will fix it.

creativeprojects:
Right, so I've been using this version for a few days now and it all looks good to me.

I think it's ready to merge 👍🏻 

Thanks again for the tremendous work! ❤️

jkellerer:
Great, I'll squash and merge now. Have using this implementation for several weeks and had no issues (but didn't use the 2 failing features 😄)

Issue #109, opened 2022-05-16
Dedicated command "generate"


jkellerer:
There are more and more things coming that generate something. In milestone 0.17.0 it is `completion-script` and the long existed `random-key`, in 0.18.0 it will be JSON schema and config file reference.

Would it make sense to use a single `generate` command to cover the different cases (like `restic` does it)?

If we implement this for 0.17.0 then we'd only have to hide and deprecate command `random-key`. If we'd wait for next milestone we'd have 2 old commands to deprecate.

creativeprojects:
yeah you're right, we should keep the commands similar to the restic ones, and it makes sense to do it now 👍🏻

jkellerer:
Ok. Will take care of it later today.

PR #110, opened 2022-05-16
Added "generate" command to create resources

jkellerer:
Implements #109 - "generate" command

Replaces commands: 
* `random-key` is hidden for backward compatibility
* `completion-script` is removed since #90 was not released yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,16 +260,16 @@ To load completions in shell or profile, use:
 
 ```shell
 # bash
-eval "$(resticprofile completion-script --bash)"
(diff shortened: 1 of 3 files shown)


jkellerer:
Codecov report is likely wrong as I had to re-run actions manually. Ready for review

creativeprojects:
Yep, looks good to me, thanks 👍🏻 

(codecov is annoying and I should try to find a replacement actually, like using the sonarcloud coverage report would be a good idea)

creativeprojects (approved):

Issue #111, opened 2022-05-26
Windows variables


tj57:
This command:

`    run-after-fail:`
`     - "curl -fsS --retry 3 --data-raw \"$ERROR_STDERR\" https://healthchecks.mydomain/ping/4b2fef02-0ccf-ff/$ERROR_EXIT_CODE"  `

works correcly under linux, but with windows not.
I tested $ERROR_STDERR, $ERROR_EXIT_CODE, %errorlevel%

Can variables in windows be the same as in linux ?

jkellerer:
resticprofile uses a shell to execute commands, therefore the commands follow the syntax of the native shell. 

In Windows env variables are referenced with `% ERROR_EXIT_CODE%` when the default shell (cmd.exe) is used. 

With Power Shell the variable syntax would be `$Env:ERROR_EXIT_CODE`.

tj57:
%ERROR_EXIT_CODE% works correclty, but %ERROR_STDERR% not.
Whether resticprofile can set the same variables on both systems ?

jkellerer:
The variables are the same but the way how they are handled may differ depending on the shell. E. g. I’m not sure if cmd.exe would tolerate variable values with newlines (like in stderr).

jkellerer:
While testing #112, I found on windows default shell (cmd.exe), double quotes are not properly escaped. You may have better results with single quotes instead. E.g.:
`"curl -fsS --retry 3 --data-raw '%ERROR_STDERR%' https://healthchecks.mydomain/ping/4b2fef02-0ccf-ff/%ERROR_EXIT_CODE%"`

But also in this case multiline is an issue. Starting a small script (that also has access to env variables) may fix this.

tj57:
Great, thank you for your time and effort.

creativeprojects:
Fixed in v0.18.0 (https://github.com/creativeprojects/resticprofile/releases/tag/v0.18.0)

PR #112, opened 2022-05-26
Allow to choose `shell` in global config

jkellerer:
In relation to #111 this PR allows to choose what shell is used in `resticprofile`:

```
.\resticprofile test-backup.snapshots

2022/05/26 15:07:15 using configuration file: profiles.yaml

global:
  shell: "powershell"

profile:
  initialize: true
  repository: local:test-repo
  password-file: test-repo.key

  run-before:
    - gc profiles.yaml
    - echo $Env:windir

  run-after-fail:
    - 'echo "EXIT: $Env:ERROR_EXIT_CODE" "STDERR: $Env:ERROR_STDERR"'

test-backup:
  inherit: profile
  backup:
    check-before: true
    check-after: true
    source:
      - test.sql

C:\Windows

2022/05/26 15:07:16 profile 'test-backup': initializing repository (if not existing)
2022/05/26 15:07:16 profile 'test-backup': starting 'snapshots'
repository 6a55e541 opened successfully, password is correct

ID        Time                 Host               Tags        Paths
---------------------------------------------------------------------------------------------------------------------------------
735928a5  2022-03-06 18:29:30  TEST                           test.sql
...
```

On error:
```
.\resticprofile test-backup.snapshots --invalid
...
2022/05/26 16:23:53 profile 'test-backup': starting 'snapshots'
unknown flag: --invalid
EXIT: 1
STDERR: unknown flag: --invalid
```

On error using default Windows shell (cmd.exe) with delayed expansion (to support newlines):
```yaml
  run-after-fail:
    - "echo 'EXIT: %ERROR_EXIT_CODE%' 'STDERR ( !ERROR_STDERR! )'"
```
results in:
```
.\resticprofile test-backup.snapshots --invalid
...
2022/05/26 17:27:43 profile 'test-backup': starting 'snapshots'
unknown flag: --invalid
'EXIT: 1' 'STDERR ( unknown flag: --invalid
 )'
2022/05/26 17:27:43 snapshots on profile 'test-backup': exit status 1
```

### DIFF
--- a/config/global.go
+++ b/config/global.go
@@ -18,6 +18,7 @@ type Global struct {
 	ResticBinary         string        `mapstructure:"restic-binary"`
 	ResticLockRetryAfter time.Duration `mapstructure:"restic-lock-retry-after"`
 	ResticStaleLockAge   time.Duration `mapstructure:"restic-stale-lock-age"`
+	ShellBinary          []string      `mapstructure:"shell"`
 	MinMemory            uint64        `mapstructure:"min-memory"`
 	Scheduler            string        `mapstructure:"scheduler"`
 	LegacyArguments      bool          `mapstructure:"legacy-arguments"` // broken arguments mode (before v0.15.0)
--- a/config/global_test.go
+++ b/config/global_test.go
@@ -49,6 +49,7 @@ nice = 1
 priority = "low"
 default-command = "version"
 initialize = true
+shell = "bash"
 restic-binary = "/tmp/restic"
 restic-lock-retry-after = "2m30s"
 restic-stale-lock-age = "4h"
@@ -65,6 +66,7 @@ restic-stale-lock-age = "4h"
 	assert.Equal(t, "low", global.Priority)
 	assert.Equal(t, "version", global.DefaultCommand)
 	assert.True(t, global.Initialize)
+	assert.Equal(t, []string{"bash"}, global.ShellBinary)
 	assert.Equal(t, "/tmp/restic", global.ResticBinary)
 	assert.Equal(t, 150*time.Second, global.ResticLockRetryAfter)
 	assert.Equal(t, 4*time.Hour, global.ResticStaleLockAge)
--- a/constants/env.go
+++ b/constants/env.go
@@ -4,6 +4,7 @@ const (
 	EnvProfileName      = "PROFILE_NAME"
 	EnvProfileCommand   = "PROFILE_COMMAND"
 	EnvError            = "ERROR"
+	EnvErrorMessage     = "ERROR_MESSAGE"
 	EnvErrorCommandLine = "ERROR_COMMANDLINE"
 	EnvErrorExitCode    = "ERROR_EXIT_CODE"
 	EnvErrorStderr      = "ERROR_STDERR"
--- a/docs/content/configuration/reference/index.md
+++ b/docs/content/configuration/reference/index.md
@@ -24,6 +24,7 @@ None of these flags are passed on the restic command line
 * **restic-lock-retry-after**: duration
 * **restic-stale-lock-age**: duration
 * **min-memory**: integer (MB) - see [memory]({{< ref "/usage/memory" >}})
+* **shell**: string (shell binary to run commands, default value is OS specific)
 * **scheduler**: string (`crond` is the only non-default value)
 * **systemd-unit-template**: string (file containing a go template to generate systemd unit file)
 * **systemd-timer-template**: string (file containing a go template to generate systemd timer file)
--- a/shell/command.go
+++ b/shell/command.go
@@ -2,10 +2,13 @@ package shell
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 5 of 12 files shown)


jkellerer:
Looking at go's `os.Expand()` implementation it looks this works only for the unix style of variables (`$var` or `${var}`). Since we have a mixture where some env vars are resolved in resticprofile (e.g. paths) and some in the shell, would it make sense to translate `${var}` to `$Env:var` and `!var!` for powershell and cmd? 

This would make sure that there is one style of env vars declaration that will always work.

As a limitation, for powershell this would mean that builtin shell variables cannot be accessed unless the translation would be something like this `$(If ($Env:var) {$Env:var} Else {$var})`.

creativeprojects:
At a glance it looks good to me. But what if the user has git_bash (or other) installed and prefers to use bash on Windows instead?

I might be old school but I was using git_bash as my default shell when I was using Windows at work, which makes configuration files more reusable in general.

I suppose I've never been a big fan of cmd or powershell anyway, so maybe it's just me

jkellerer:
Was also thinking on non common options like bash on windows and powershell on Unix. Will update the PR to handle all these cases.

jkellerer:
The latest update decouples OS and shell and translates $var to native shell env vars syntax (for cmd.exe and powershell).

E.g. powershell on MacOS works the same as on Windows. Also the same profile works fine bash as (simple) commands like `echo $PROFILE_NAME` or `echo $ERROR_STDERR` work the same on every shell:

MacOS

```
2022/06/05 16:34:53 profile 'test-backup': initializing repository (if not existing)
2022/06/05 16:34:53 starting command: /usr/local/bin/restic init --password-file test-repo.key --repo local:test-repo
2022/06/05 16:34:53 dry-run: /usr/local/bin/restic init --password-file test-repo.key --repo local:test-repo
2022/06/05 16:34:53 dry-run shell: /usr/local/bin/pwsh -Command /usr/local/bin/restic init --password-file test-repo.key --repo local:test-repo
2022/06/05 16:34:53 starting pre-backup command 1/1
2022/06/05 16:34:53 dry-run: echo "Backup $PROFILE_NAME" 
2022/06/05 16:34:53 dry-run shell: /usr/local/bin/pwsh -Command echo "Backup $($_=${Env:PROFILE_NAME}; if ($_) {$_} else {${PROFILE_NAME}})"
```

Windows

```
2022/06/05 16:36:44 profile 'test-backup': initializing repository (if not existing)
2022/06/05 16:36:44 starting command: restic.exe init --password-file test-repo.key --repo local:test-repo
2022/06/05 16:36:44 dry-run: restic.exe init --password-file test-repo.key --repo local:test-repo
2022/06/05 16:36:44 dry-run shell: C:\Program Files\PowerShell\7\pwsh.exe -Command .\restic.exe init --password-file test-repo.key --repo local:test-repo
2022/06/05 16:36:44 starting pre-backup command 1/1
2022/06/05 16:36:44 dry-run: echo "Backup $PROFILE_NAME"
2022/06/05 16:36:44 dry-run shell: C:\Program Files\PowerShell\7\pwsh.exe -Command echo "Backup $($_=${Env:PROFILE_NAME}; if ($_) {$_} else {${PROFILE_NAME}})"
```

This allows to write profiles that work on any OS (to some extend if pre/post commands are generic enough to be compatible)

In some cases powershell does behave differently, had tried to use `pwsh` as default on Windows where it is installed (as `pwsh` is always a manual install), but as github actions have it installed it broke several tests and at least the test for NoErrorOnWarning was not fixable as pwsh seems to not return the error code properly for the test so I reverted the changes and default for Windows remains `cmd.exe` in all cases.

creativeprojects:
At first glance it looks pretty good.

I tried a few of my exotic configurations on my Windows 10 VM, and this one is puzzling me:

This is the configuration file
```yaml
stdin-command:
    inherit: default
    backup:
        stdin-command:
          - "type ${USERPROFILE}/NTUSER.DAT"
        stdin-filename: stdin-test

```

The normal Windows way `%USERPROFILE%` doesn't get replaced and stays as is.

Then I tried with a `/`, `\` and `\\` separators and they all give me the same cryptic error:

```
Get-Content : A positional parameter cannot be found that accepts argument '/NTUSER.DAT'.
At line:1 char:1
+ type $($_=${Env:USERPROFILE}; if ($_) {$_} else {${USERPROFILE}})/NTU ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Get-Content], ParameterBindingException
    + FullyQualifiedErrorId : PositionalParameterNotFound,Microsoft.PowerShell.Commands.GetContentCommand
```

Does that make any sense to you?

jkellerer:
Thanks for testing this one, had overlooked this part.

After checking it in the shell, it turns out that `$(...)/path` in PS is not the same as `${VAR}/path`, for the latter the whole evaluates to a string (even without quotes) for the former, `$()` evaluates first and is treated as a separate token leading to 2 parameters for `Get-Content` instead of a single path.

This means the following:
```
type $($_=${Env:USERPROFILE}; if ($_) {$_} else {${USERPROFILE}})/NTUSER.DAT
Get-Content : A positional parameter cannot be found that accepts argument '/NTUSER.DAT'.
...

type "$($_=${Env:USERPROFILE}; if ($_) {$_} else {${USERPROFILE}})/NTUSER.DAT"
type : The process cannot access the file 'C:\Users\guest\NTUSER.DAT' because it is being used by another process.
...

type ${Env:USERPROFILE}/NTUSER.DAT
type : The process cannot access the file 'C:\Users\guest\NTUSER.DAT' because it is being used by another process.
...
```

While it is in general a good practice to quote paths, I guess the expression $(...) cannot be used in this case as it changes default behavior. On the other hand, I've create a list of all built-in vars of PS (using MS documentation on PS7). So we can use the simple replacement `${Env:var}` for all others and if it turns out to be an issue still add a switch to disable the feature later.

creativeprojects (approved):
Looks good to me 👍🏻

PR #114, opened 2022-05-26
Add http hooks

### LINKED ISSUE #102: Healthchecks integration
Hi,
I have a feature request: An Integration of Healthchecks. There is a [hosted version](https://healthchecks.io/), but it can also be [selfhosted](https://github.com/healthchecks/healthchecks). It should do the following things:

1. Webhook when starting a location
2. Webhook when finishing a location, including status (or success/fail) as well as a (truncated?) log.

This would be similar to e.g., [borgmatic](https://torsion.org/borgmatic/docs/how-to/monitor-your-backups/#healthchecks-hook). The benefit of having that directly integrated here is that it is one thing less to break. Or in a Windows case, one thing less to properly install [curl] as well as not having to fight with questionable Windows batch syntax just to set up a backup.

creativeprojects:
HTTP hooks work the same way the command hooks are running.

The sections that allow sending HTTP hooks are:
- backup
- copy
- check
- forget
- prune

You might notice that's also the same sections that can be scheduled

Each of these commands can send 4 different types of hooks:

- send-before
- send-after
- send-after-fail
- send-finally

The configuration is the same for each of these 4 types of hooks:

| Name | Required | Default | Notes |
|:-----|:---------|:--------|:------|
| url | Yes | None | URL of your Webhook |
| method | No | GET | this is the HTTP method (GET, POST, HEAD, etc.) |
| skip-tls-verification | No | False | Use only if you're using your own server with a self-signed certificate: **This is not recommended** |
| headers | No | User-Agent set to resticprofile | This is a subsection with a list of `name` and `value` |
| body | No | Empty | Used to send data tp the Webhook (POST, PUT, PATCH) |
| body-template | No | None | Template file to generate the body (in go template format) |

Please note a fail during a hook **will not fail the job**. My take on it is that failing to send monitoring information shouldn't fail the whole backup script.
If you need this functionality I can add a flag for it.

Here's an example on how to use it:

```yaml
profile:
    inherit: default

    backup:
        source: "/source"
        exclude:
          - "/**/.git/"
        schedule:
          - "*:00,30"
        schedule-permission: user

        # you can have more than one target
        send-before:
          - method: HEAD
            url: https://hc-ping.com/831e288e-1293-46f8-ac31-70ea7f875650/start
          - method: HEAD
            url: https://httpstat.us/400

        # you can have more than one target
        send-after:
          - method: HEAD
            url: https://hc-ping.com/831e288e-1293-46f8-ac31-70ea7f875650
          - method: HEAD
            url: https://httpstat.us/500

        send-after-fail:
            method: POST
            url: https://hc-ping.com/831e288e-1293-46f8-ac31-70ea7f875650/fail
            body: "${ERROR}\n\n${ERROR_STDERR}"
            headers:
              - name: Content-Type
                value: "text/plain; charset=UTF-8"
    check:
        schedule:
          - "*:15"

        send-before:
          method: HEAD
          url: https://hc-ping.com/e0f62e41-b75f-450f-8cdd-7f25e466d2dc/start

        send-after:
          method: HEAD
          url: https://hc-ping.com/e0f62e41-b75f-450f-8cdd-7f25e466d2dc

    retention:
        after-backup: true
```

I just need to update the documentation and it's ready 👍🏻 

Fixes #102


creativeprojects:
> * For diagnostics it would be good to log the full response (e.g. basic info like return code and message as debug when 20x and warning for other codes, headers and body as trace). I think err is only set for serious errors not for http codes other than 20x.

Yes I was thinking about it but It can be added later

> * The usual SSL options cannot be set (ca-cert, client-cert, insecure-option).
> * When timeout is hard coded we might use 30 or 60 seconds? Or add a config option.

I knew I was missing something but I couldn't remember what it was 😄

Issue #80, opened 2021-10-19
[proposal] Improved configuration file format


creativeprojects:
# Proposal for an updated configuration file format

## Introduction

The current file format was decided at the time resticprofile was only using the `toml` format. Nesting pieces of configuration in blocks is not the easiest as you have to specify the whole path in the block:

```toml
[profile]

[profile.backup]
...

```

Since then, I believe the `yaml` format is preferred over `toml`.

My proposal is to make a `version 2` of the configuration file, the current file format is `version 1`.

**Both formats will continue to be valid** (like `docker-compose`):

- if no version is specified, the `version 1` is used. This is the current format
- if a version is specified (`2`) the new format will be expected

## New format availability

The new format `version 2` will be available for:
- TOML
- YAML
- JSON

**It won't be available for HCL**. This may not be definitive, but it's not widely used and it's becoming more and more difficult to support HCL.

**HCL can still be used as is, `version = 1`**

## New format specifications

I will show the specification using the `yaml` as examples, because it's probably the most readable format.

### version

```yaml
---
version: 2
```

### global

The global section does not change. We'll keep all the global configuration in there.

```yaml
---
global:
    default-command: snapshots
    initialize: false
    priority: low

```

### profiles

All your profiles will be nested under a `profiles` section. Please note the schedules are no longer described inside the profile, but in a separate section `schedules` (see following sections).

```yaml
profiles:
    default:
        env:
            tmp: /tmp
        password-file: key
        repository: /backup

    documents:
        inherit: default
        backup:
            source: ~/Documents
        snapshots:
            tag:
                - documents
```

### groups

The list of profiles will be nested under a `profiles` section, so we can add more configuration to groups later.

```yaml
groups:
    full: # name of your group
        profiles:
            - root
            - documents
            - mysql
```

This format leaves more space for improvements later (like a `repos` section maybe?)

Does that make sense to you?

Please reply in this issue, thanks :)

*Edit*:
- changed numbering as such: `1` is current and `2` is the new format
- HCL not supported with version 2
- `schedules` section removed

creativeprojects:
As a matter of fact, this is the reason why I need to modernise the configuration, to allow for a `targets` section 😛 

(I think you should be able to update the Wiki now?)

jkellerer:
Yes it works. Try to contribute asap.

creativeprojects:
Using go templates is a bit cumbersome, specially for non developers who never used a similar templating language...

[...]

PR #113, opened 2022-05-26
ConfigV2: Replace list params - fix #108

jkellerer:
This is a fix for #108 in version 2 configs (version 2 configs #80 are available but not yet in the documentation). Version 1 configs keep the unintuitive behaviour to avoid breaking any existing profiles.

In order to extend list params with inheritance, an additional PR may provide extended config syntax for this (is covered in #115).

PR #115, opened 2022-05-28
ConfigV2: Mixins

jkellerer:
Provides configuration `mixins` (multiple inheritance). See also "templates" in https://github.com/creativeprojects/resticprofile/issues/80#issuecomment-1140060881

Status
- [x] Implementation
- [x] Tests
- [x] Documentation
- [x] Review of details (e.g. parameter append syntax, variable syntax, default-vars)
- [ ] Code review

Example:
```yaml
version: 2

mixins:
  healthchecks:
    default-vars:
        URL: https://hc-ping/uuid

    send-before:
        method: HEAD
        url: ${URL}/start

    send-after:
        method: HEAD
        url: ${URL}

    send-after-fail:
        method: POST
        url: ${URL}/fail
        body: "${ERROR}\n\n${ERROR_STDERR}"
        headers:
          - name: Content-Type
            value: "text/plain; charset=UTF-8"

profiles:
  # simple syntax
  with_monitoring:
    use:
      - healthchecks
    backup: 
      source: /

  # full syntax
  with_monitoring_and_uuid:
    use:
      - name: healthchecks
        vars:
          URL:  https://hc-ping/another-uuid
    backup: 
      source: /
```

List append example ("key..." appends to a list, "...key" prepends to a list)
```yaml
version: 2

mixins:
  vm-snapshot:
    default-vars:
        LIVEDATA: /var/db
    
    ...run-before:
      - virsh snapshot-create-as --domain "${VM}" --name "${VM}-backup" --diskspec "vda,file=${LIVEDATA}/${VM}-livedata.qcow2" --disk-only --atomic --no-metadata --quiesce
      
    run-finally...:
      - [ -e "${LIVEDATA}/${VM}-livedata.qcow2" ] && virsh blockcommit --domain "${VM}" vda --wait --active && virsh blockjob --domain "${VM}" "${LIVEDATA}/${VM}-livedata.qcow2" --pivot && rm -f "${LIVEDATA}/${VM}-livedata.qcow2"

profiles:
  vms:
    backup:
      use:
        - {name: vm-snapshot, VM: my-app-server}
        - {name: vm-snapshot, VM: my-docker-server}

      source:
        - /opt/vms/my-app-server*
        - /opt/vms/my-docker-server*

      run-before:
        - echo all snapshots created
      run-after:
        - echo all vm snapshots backed-up
      run-finally:
        - echo removing all existing snapshots now
```


jkellerer:
Hi @creativeprojects , would you mind reviewing the current implementation and concept whether this meets your expectations.

Once you are fine, I'll check how I can add the changes to the documentation and after that I'll mark the PR as ready.

creativeprojects:
It looks good to me so far 👍🏻 

I was thinking about using `<` and `>` instead of `+` but it's not allowed on TOML so far. I haven't checked but it's probably not allowed on the other formats either.

Would that make it more readable to put it in the values instead?

like ["++", "other value"] or ["other value", "++"]? Just thinking out loud here.

creativeprojects:
If the default tags are messing up with some of the user's values, we could make both tags configurable in the global section though?

jkellerer:
### Suffixes for appending

While I think in general it is a good idea to make things configurable, for the suffixes (or tags) we probably should decide on a syntax to simplify documentation and sharing of snippets.

Putting append or prepend logic in the value is an option but I think it would confuse me a bit in a config file, might just be me though.

TOML officially allows the following chars `[A-Za-z0-9_-]` in config keys (unless they are quoted).

Only valid in YAML, TOML requires quoted key for the following variations:

```plain
property>        run-after-fail>: ...
property<        run-after-fail<: ...

property>>       run-after-fail>>: ...
property<<       run-after-fail<<: ...

property+>       run-after-fail+>: ...
property->       run-after-fail->: ...

property++       run-after-fail++: ...
property+0       run-after-fail+0: ...

property__++     run-after-fail__++: ...
property__0+     run-after-fail__0+: ...

property+LAST    run-after-fail+LAST: ...
property+FIRST   run-after-fail+FIRST: ...
```

Valid for any config file format:

```plain
property__LAST         run-after-fail__LAST: ...
property__FIRST        run-after-fail__FIRST: ...

property__APPEND       run-after-fail__APPEND: ...
property__PREPEND      run-after-fail__PREPEND: ...

APPEND__property     APPEND__run-after-fail: ...
PREPEND__property    PREPEND__run-after-fail: ...
```

I would tend to `property__APPEND` / `property__PREPEND` and maybe allow an alternate short syntax from the list above (e.g. `property__++` / `property__0+`). There should be no overlaps with existing config options and also this is only allowed within templates, not within other areas of the configuration.

### Naming it

At several places we already use "template" and mean go templates for something (mostly to generate markup of some sort). It probably may confuse users when this feature is also called "templates". I've used the term "inline-template" as a working title to have something for coding but it is probably not too good.

Logically this feature is more related to `mixins` (multiple inheritance) and/or config `snippets`, how about:

```yaml
version: 2

mixins:
  exclude-hidden:
    backup:
      exclude__APPEND: [".*", "*~"]

profiles:
  projects:
    use: ["exclude-hidden"]
    backup:
      source: /projects/src
```

creativeprojects:
Thanks for your research, this is very much appreciated 🙇🏻 

I do like the `PREPEND` and `APPEND`. This is easy to remember, and easy to read. I can't really decide before or after the field name, maybe both? it's just a case of `HasPrefix || HasSuffix` after all.

I guess it depends how you're going to read it in your head:
- append to this field using these values
- take this field and append these values

I thought about another cool one, which works in YAML (but sadly not TOML):
`...for_prepend` and `for_append...`

I guess we could allow it on YAML and only the APPEND/PREPEND on TOML? maybe

---

Now your idea of `mixins` and `use` is spot on!
I also had concerns about providing two different ways of doing templating, and that closes the case very nicely 😄

jkellerer:
Successfully renamed to `mixins` and changed syntax to `__APPEND` (or `key...`) and `__PREPEND` (or `...key`). Even the following toml works fine:
```toml
[mixins.exclude-macos]
"exclude..." = [
    ".DS_Store",
]
```

You might notice that I made key-delimiter as used by `viper` configurable at locations where it wasn't. It is "." and should conflict with the short syntax but surprisingly, it works without changing it. Should it break with an update, then it can be changed and tests will still succeed.

Next is documentation. Think I'll start with a new page for "inheritance" that explains "inherit", "mixin" and how flags are passed from top to bottom as I this topic isn't really covered so far on its own?

creativeprojects:
I'm amazed the 3 dots work with TOML, that's excellent news.

For the documentation, I'm afraid there's nothing for inherit. I started this project on my own requirements and since there were plenty similar ones I didn't bother much with documentation. Inherit was implemented from the beginning.

So I was just relying on the user reading a big configuration file to understand it all 😄 :
https://creativeprojects.github.io/resticprofile/configuration/examples/

jkellerer:
@creativeprojects , PR is awaiting a review :)

creativeprojects:
Cool, I'll have a look 👍🏻 

I approved #113 but I don't know if you want to wait for this one

jkellerer:
The 2 PRs should not have conflicts and can be merged in any order (though the documentation enhancements in this PR assume that #113 is already merged).

creativeprojects:
Right, I'll merge #113 now then 😉

jkellerer:
Looks like it is still valid, will rebase and run tests just to be sure 😀

docs/content/configuration/inheritance/_index.md
```diff
@@ -0,0 +1,425 @@
...
+weight: 16
+---
+
+{{% notice style="tip" %}}
+You can use `resticprofile [<profile-name>.]show` to see the effect inheritance on a profile
+{{% /notice %}}
+
+## Profile Inheritance
+
+Profiles can inherit from a parent profile. This allows to define the general behavior and common configuration in a base profile while **derived** profiles only define what is specific, e.g. what needs to be included in the backup or which command [hooks](../hooks/) (e.g. `run-before`, `run-after` & `run-finally`) must be started.
```
creativeprojects, line 13: you can use the macro {{< ref "/configuration/hooks" >}} to generate a url. The good thing is that it's going to fail when we rename a chapter 😉

docs/content/configuration/inheritance/_index.md
```diff
@@ -0,0 +1,425 @@
...
+
+When assuming profile "*backup-homes*" inherits from profile "*base*", then the effective configruation of "*backup-homes*" is built by merging the profile configuration of "*backup-homes*" into "*base*".
```
creativeprojects, line 15: typo: configuration

docs/content/configuration/include/_index.md
```diff
@@ -70,4 +70,117 @@ global {
...
+/etc/resticprofile/conf.d/01_b.conf
+/etc/resticprofile/last.conf
+```
+
+Configuration **merging** follows the logic:
+
+* Configuration properties are replaced
+* Configuration structure (tree) is merged
+* What includes later overrides what defines earlier
+* Lists of values or objects (tree structure) are considered properties not config structure and will be replaced
```
creativeprojects, line 102: as such it's not part of the "merging" process, but should we also note where the go template gets resolved in the grand scheme of things?

creativeprojects (commented):
It all looks good to me 😄 

I tried a few mixins and they all got resolved as expected.

I just have a small comment: should we hide the "use" section in the `show` command? since the profile has been merged at that point.

Nice work 👍🏻

docs/content/configuration/inheritance/_index.md
jkellerer: Was using IDE auto-complete but yeah, compile time check makes sense 😀

docs/content/configuration/include/_index.md
jkellerer: Makes perfect sense to me. Added it as introduction to "loading" as it's important that this is understood.

jkellerer:
Regarding hiding `use`: Had first set the use property to nil to hide it by removing it (`viper` api allows setting but not removing properties). But viper has very strange behaviour when setting `some.path.property`. This sets `some` and `some.path` to an empty map so setting `profiles.profile.use` to `nil` effectively also sets `profile` to undefined.

It could be removed in `show` and this would be consistent with `inherit`. Hmm, probably better to be consistent.

creativeprojects (approved):
Great work!

Now, if I remember well, the only thing missing for v2 is the scheduling (since I decided to move it to its own section). And then I need to go round everything to make sure I haven't forgotten some bits...

Thanks 😉

Issue #105, opened 2022-04-24
Remote syslog


bdillahu:
Maybe I'm missing the obvious way of doing it, but any thoughts of supporting logging to a remote syslog server in addition to the log file methods?

creativeprojects:
Yeah I can add that on my list. That's easy enough to implement (I've done it before).

creativeprojects:
Do you just want the resticprofile logs sent to syslog right?

There's no need to capture the restic output to send to syslog as well?

bdillahu:
Correct, just the resticprofile logs were my thought...

PR #127, opened 2022-07-23
Add support for syslog

creativeprojects:
This PR adds the ability to specify a syslog server as a `log` target. It works for:

- the `--log` option on the command line
- the `schedule-log` flag in the configuration file (on all targets available for scheduling)
- 
The format of the server url is `proto://server_ip:server_port`, like `tcp://localhost:514`

Fixes https://github.com/creativeprojects/resticprofile/issues/105


jkellerer:
Hi @creativeprojects, as far as I can see `syslog` and `logfile` cannot be used in parallel (which is totally fine). 

Was wondering whether it would make sense to retain only one "log" option instead of "log", "logfile" and "syslog", where the single "log" option  supports logging like this:

```
resticprofile --log /path/to/logfile.log         # setupFileLogger with logfile.log
resticprofile --log C:/path/to/logfile.log       # setupFileLogger with logfile.log
resticprofile --log syslog                       # setupSyslogLogger with local syslog
resticprofile --log syslog-tcp://hostname(:port) # setupSyslogLogger with tcp remote log
resticprofile --log syslog-udp://hostname(:port) # setupSyslogLogger with udp remote log
resticprofile --log syslog://hostname(:port)     # setupSyslogLogger with udp remote log
```

This prevents that --syslog and --log can be used in parallel (which would have not the expected effect) and it does not require changes to scheduling.

Also one other thought for Windows: I think the error should be printed when syslog is attempted to be used, so that it is clear that for this platform syslog is not supported.

creativeprojects:
Yes, this is a very good point 👍🏻 

Thanks for your feedback, it is much appreciated 😄

creativeprojects:
@jkellerer it turns out to be a bigger refactoring than initially thought 😛 

I've removed the `JobConfig` interface which wasn't very useful after the refactoring I've done on the scheduler some time ago (by using a `Handler`).

I've introduced the use of the mock library from testify

The good news is that now we can test the functions in `schedule_jobs.go`. More to come but it's out of the scope of this PR.

jkellerer:
Will have a look at it later today.

jkellerer:
Looks good me, the simplification makes sense (removal of interfaces). Also the addition of mock lib is nice.

creativeprojects:
Cool! Thanks 😄

PR #139, opened 2022-08-28
Add --help to own commands (e.g. generate --help)

jkellerer:
This PR adds a new `help` command supporting the following:

| Commandline                             | Purpose                                          |
|-----------------------------------------|--------------------------------------------------|
| `resticprofile OWNCOMMAND --help` or `resticprofile help OWNCOMMAND` | Display detailed command help including supported flags |
| `resticprofile help COMMAND`            | Display restic command help with extra information what flags are set by profiles |
| `resticprofile help`                    | Same as `resticprofile --help`                   |

E.g.:

```
➜ resticprofile help version

The "version" command displays brief or detailed version information

Usage:
   resticprofile [resticprofile flags] version [command specific flags]

Flags:
   -v, --verbose   display detailed version information
```

---

```
➜ resticprofile help prune 

The "prune" command checks the repository and removes data that is not
referenced and therefore not needed any more.

EXIT STATUS
===========

Exit status is 0 if the command was successful, and non-zero if there was any error.

Usage:
  resticprofile [resticprofile flags] [profile name.]prune [flags]

Flags:
  -n, --dry-run                               do not modify the repository, just print what would be done
  -h, --help                                  help for prune

....

Flags applied by resticprofile (configuration "profiles.yaml"):

   profile "documents":
      --password-file   test-repo.key     
      --repo            local:test-repo   
   profile "p1-base":
      --password-file   test-repo.key     
      --repo            local:test-repo   
   profile "p1-overridden":
      --password-file   test-repo.key     
      --repo            local:test-repo   

....
```

Besides the things mentioned above, the `help` command also introduces slight use of ANSI colors and respects the width of the console window.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -10,11 +10,9 @@ import (
 	"os"
 	"regexp"
 	"runtime"
-	"runtime/debug"
 	"sort"
 	"strconv"
 	"strings"
-	"text/tabwriter"
 
 	"github.com/creativeprojects/clog"
 	"github.com/creativeprojects/resticprofile/config"
@@ -28,6 +26,7 @@ import (
 type ownCommand struct {
 	name              string
 	description       string
+	longDescription   string
 	action            func(io.Writer, *config.Config, commandLineFlags, []string) error
 	needConfiguration bool              // true if the action needs a configuration file loaded
 	hide              bool              // don't display the command in help and completion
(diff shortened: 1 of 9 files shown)


creativeprojects:
Yes this a good idea.

We could also support `resticprofile help generate` since it is a rather common pattern as well as `resticprofile generate -h`

... and maybe we should also permit to pass the help through to restic: for now a `resticprofile backup -h` is trying to load a `default` profile before calling restic, and fails if it didn't find a configuration file.

jkellerer:
I’ll update the PR this week. 

If we add a `help` command that also forwards to restic help, I think we could display what flags the different profiles define (if any), so that the normal help of restic is extended with the config that resticprofile applies.

creativeprojects:
Oh yes, good idea, I do like that 👍🏻

jkellerer:
@creativeprojects, the last week in Aug was longer than expected :) ... let me know what you think.

creativeprojects:
> the last week in Aug was longer than expected

I know that feeling 😄 

So far it looks good to me.
I noticed that we need to escape the `%` though, if you try a `help prune` you get a `(MISSING)` after a `%`:

```
% ./resticprofile help prune

The "prune" command checks the repository and removes data that is not
referenced and therefore not needed any more.

EXIT STATUS
===========

Exit status is 0 if the command was successful, and non-zero if there was any error.

Usage:
  resticprofile [resticprofile flags] [profile name.]prune [flags]

Flags:
  -n, --dry-run                               do not modify the repository, just print what would be done
  -h, --help                                  help for prune
      --max-repack-size size                  maximum size to repack (allowed suffixes: k/K, m/M, g/G, t/T)
      --max-unused limit                      tolerate given limit of unused data (absolute value in bytes with suffixes k/K, m/M, g/G, t/T, a value in
                                              %!o(MISSING)r the word 'unlimited') (default "5%!"(MISSING))
      --repack-cacheable-only                 only repack packs which are cacheable
      --repack-small                          repack pack files below 80%!o(MISSING)f target pack size
      --repack-uncompressed                   repack all uncompressed data
      --unsafe-recover-no-free-space string   UNSAFE, READ THE DOCUMENTATION BEFORE USING! Try to recover a repository stuck with no free space. Do not use
                                              without trying out 'prune --max-repack-size 0' first.

...
```

jkellerer:
Have added the missing unit test. Should be fixed now.

creativeprojects (approved):
It looks good to me 👍🏻 

Thanks again for your `help` 😄 

and that will make it to 0.19.0

PR #146, opened 2022-10-27
New schedule section for configuration v2

creativeprojects:
This PR is adding support for the `schedules` section on the configuration format v2:

```yaml
schedules:
    full-backup: # give a name to your schedule
        group: full
        schedule:
            - "Mon..Fri *:00,15,30,45" # every 15 minutes on weekdays
        permission: user
        run: backup # backup is the default if not specified

    other:
        profiles:
            - root
            - mysql
        schedule:
            - "Sat,Sun 0,12:00" # twice a day on week-ends
        permission: user
        run: prune
```

- [ ] Implementation
- [ ] Tests
- [ ] Documentation
- [ ] Ready for review

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -552,7 +552,13 @@ func getScheduleJobs(c *config.Config, flags commandLineFlags) (schedule.Schedul
 		return nil, nil, nil, fmt.Errorf("cannot load profile '%s': %w", flags.name, err)
 	}
 
-	return schedule.NewSchedulerConfig(global), profile, profile.Schedules(), nil
+	schedules := profile.Schedules()
+	// schedulesConfig := make([]*config.ScheduleConfig, len(schedules))
+	// for index, schedule := range schedules {
+	// 	schedulesConfig[index] = schedule.GetScheduleConfig()
+	// }
+
+	return schedule.NewSchedulerConfig(global), profile, schedules, nil
 }
 
 func requireScheduleJobs(schedules []*config.Schedule, flags commandLineFlags) error {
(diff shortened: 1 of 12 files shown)


creativeprojects:
I finally got a bit of time to do some work on the new scheduling configuration.

@jkellerer I wanted to discuss a few options before I go further. The way we call resticprofile from the scheduler will have to change. That's because we allow a list of profiles to be part of the schedule (which I think is a good thing anyway).

From now I can see two options:
- add a `--profiles profile1,profile1` option. This is the closest to the current implementation. We can change the name of the schedule, change most the schedule configuration and it's still going to work (the schedule configuration is transformed into command line flags). Except for the list of profiles. If we change the list of profile we have to `unschedule` and `schedule` again.
- add a `--run-schedule name` option. With this option we don't need to transform the configuration into flags. We just need to read the schedule section when we start the job. This is a bit different than what we do now but it would make it much easier to change the configuration without rescheduling the jobs. One drawback though is that we cannot change the name of the schedule.

By writing it I realise I like the second option a bit more, but I wanted to have your opinion on it 😉

jkellerer:
Reading it, intuitively, I also like option 2 more (before reading the last line 😀). So looks like we're on the same page.

jkellerer:
Regarding the name.. I was thinking that it might be something to consider to have kind of cache and/or state storage which might be used for all sorts of things including detecting name changes. 

While it may not help now, we could use it later to automatically act when name changes are detected (not only in this case).

creativeprojects:
Replaced by #333

PR #149, opened 2022-11-07
Continue profile in group on error

### LINKED ISSUE #135: [Feature request] Add flag to continue group backup if one or more profiles return with error
Hello,
I'm trying to using resticprofile to back up some VMs, but this can only happen, if a VM is currently not running. Therefore I made a profile for each VM and added all profiles to a group, so I can back them up with one command.

To check if a VM is currently running I used the run-before backup command hook on each profile. A small script then checked the VM state and returned a non-zero exit code, if the VM is running. My hope was that only the corresponding profile would get skipped and resticprofile would continue with the next one, so all non-running VMs would get backed up, whilst skipping the other. However when a command hook returns a non-zero exit code the complete action is cancelled and all remaining profiles are not processed.

Suggestion: Because the new config file format #80 already was prepared to allow for more configuration of groups I think a flag like "on-error-continue-with-next" might be relatively easy to implement?

Thanks for the awesome work so far 👍

creativeprojects:
- add one global flag `group-continue-on-error` in configuration file format `v1`
- add one group flag `continue-on-error` in configuration format `v2`

When setting the global flag `group-continue-on-error` to true, it allows the next profile(s) to run after a failing profile. Without the option all the following profiles are not started.

In a similar way on configuration file `v2` you can set the `continue-on-error` on each group independently:

```yaml
groups:
  double:
    continue-on-error: true
    profiles:
      - profile1
      - profile2

profiles:
  profile1:

  profile2:

```

Fixes #135

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,15 +23,15 @@ jobs:
 
     steps:
 
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v3
+
       - name: Set up Go ${{ matrix.go_version }}
         uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go_version }}
           check-latest: true
           cache: true
-
-      - name: Check out code into the Go module directory
-        uses: actions/checkout@v3
 
       - name: Generate mocks
         run: make mocks
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,8 @@ jobs:
         uses: actions/setup-go@v3
         with:
           go-version: ${{ matrix.go_version }}
+          check-latest: true
+          cache: true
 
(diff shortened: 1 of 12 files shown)


jkellerer:
Does it make sense to use 3 state bool at group level? 

That would allow to set it explicitly to false at group level while enabling it globally.

creativeprojects:
Yeah I was thinking about it, I suppose it doesn't hurt.

My reasoning was that the global flag wouldn't be used in `v2` but I guess it can be

creativeprojects:
Hmmm that will look funny on TOML actually:

```toml
# bool flag
flag = true
# 3 settings bool will be a string
flag = "true"
```

creativeprojects:
Actually I can just use a pointer to a bool. I don't particularly like it but it would work

PR #150, opened 2022-11-08
Fix zsh completion script (use of outdated CLI)

jkellerer:


jkellerer:
@creativeprojects , will merge it right away, as it is clearly a bug.

Issue #151, opened 2022-11-18
Add command hooks for 'restore' and 'dump' commands


ozboss:
Please add support for command hooks for the 'restore' and 'dump' commands. My intention is to stop a program before restoring its data and then start the program again. In the process some specific steps are required, so that I cannot use the global command hooks.

PR #154, opened 2022-11-18
Run-hooks for all non-conflicting commands (#151)

jkellerer:
Adds shell command hooks to the following additional commands: `dump`, `find`, `ls`, `mount`, `restore`, `snapshots`, `stats` and `tag`.

This is an enhancement to #138 and implements #151 

The remaining commands `check`, `forget`, `init`, `retention` and `prune` depend on the _alias_ feature discussed in #138 and should be added when aliases are implemented.


creativeprojects (approved):
LGTM 👍🏻

PR #158, opened 2023-01-19
Update Docker to include SSH and latest binaries

nicgord:
SFTP repositories not working in Docker image as SSH is not installed. Proposing to install SSH in contaimner.
This is tested and works well, need to include instructions to mount .ssh as a volume like this 
   volumes:
      - ~/.ssh:/root/.ssh

Also updated the Dockerfile to pull the latest versions of restic and rclone  (and optionally resticprofile) instead of using a static local binary.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -4,9 +4,21 @@ LABEL maintainer Fred <fred@gcreativeprojects.tech>
 
 ARG ARCH=amd64
 
-COPY build/restic-${ARCH} /usr/bin/restic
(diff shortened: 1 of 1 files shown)


nicgord:
Thanks for the quick response, will be great to have SSH integrated and so long as restic is latest doesn't matter how it gets in there.

Maybe consider also adding curl while you are there? May be useful for those that would prefer to use instead of HHTP Hooks as has additional options or may be more familier for some.

creativeprojects:
I've opened a new PR with ssh client, curl (and tzdata) in the docker image  #161

PR #160, opened 2023-01-20
add suppport for user_logged_on

### LINKED ISSUE #155: Task Scheduler: Run only when user is logged on
Is it possible to add the option to schedule on windows with toe option to "Run only when user is logged on"?
Otherwise it will fail for users that don't have the policy setting for [logging on as a batch job](https://learn.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/log-on-as-a-batch-job).

creativeprojects:
Adds a new option to `schedule-permission` **for Windows only**
- `system` run as built-in user `SYSTEM` - **same as before**
- `user` run as current user (logged on or not) - **same as before**
- `user_logged_on` run when current user is logged on (no need for a password in that case)

Fixes #155

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -141,7 +141,7 @@ func getOwnCommands() []ownCommand {
 			name:              "elevation",
 			description:       "test windows elevated mode",
 			action:            testElevationCommand,
-			needConfiguration: true,
+			needConfiguration: false,
 			hide:              true,
 		},
 		{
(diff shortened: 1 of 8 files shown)


jkellerer:
👍

Small side note: Correct me if I'm wrong, but `user_logged_on` is also the behaviour of systemd when run as `user`? Just wondering if we should keep it more general and say that some schedulers can support `user_logged_on` only and that `user` behaves the same in this case? That would mean that `user_logged_on` is not a Windows only option, but is an option not supported by all schedulers and implicit for others (systemd).

creativeprojects:
hmmm 🤔 that's a good question, I don't normally use schedules in `user` mode so I can't remember.
Thinking of which, it could be the same for launchd actually, I need to look it up.

Thanks for pointing it out 👍🏻

PR #161, opened 2023-01-21
New docker image

### LINKED ISSUE #159: Add tzdata package to docker image
Please add the package tzdata to your Dockerfile to let the user change the timezone.
This would be useful if you want to run crond inside Docker and to get correct time in the log files.

Example:
```
FROM alpine:latest
RUN apk add --no-cache tzdata
ENV TZ=Etc/UTC
```

creativeprojects:
Docker image with these new packages added:

- ssh client (needed for `sftp`)
- tzdata (to get local time in logs)
- curl (has more options than resticprofile hooks)
- ca-certificates

Fixes #159 
Closes #158

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -9,7 +9,7 @@ COPY build/restic-${ARCH} /usr/bin/restic
 COPY build/rclone-${ARCH} /usr/bin/rclone
 COPY resticprofile /usr/bin/resticprofile
 
-RUN apk add --no-cache openssh-client-default curl tzdata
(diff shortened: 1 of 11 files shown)


jkellerer:
@creativeprojects : Would you mind adding "ca-certificates"?

creativeprojects:
> @creativeprojects : Would you mind adding "ca-certificates"?

Sure. As it turns out, I don't normally use Alpine as a base image...

jkellerer:
Yes it's quite tiny :) but that's also the advantage

PR #172, opened 2023-02-09
Allow config encoding in UTF8-BOM, UTF16 and ISO88591

jkellerer:
This PR allows config files to be encoded with BOM (UTF8/UTF16) or in ISO-8859-1 8bit charset.

Found when testing in Windows, where UTF16 or ISO-8859-1 is more common and even UTF8 may get a BOM and fails.


creativeprojects (approved):
Well spotted!

Thanks for the fix, it looks good to me 👍🏻

Issue #168, opened 2023-01-27
Issues sending a restic log to the Windows event view


Schroedingers-Cat:
I'm trying to send restic's output log to the windows event view via `run-after` using a custom powershell function. I'm invoking this via `run-after  = "powershell -command <my-powershell-command with params>"`.  

This was tricky to get working because:
- during `run-after` there's no variable available containing restic's log (as opposed to `run-after-fail` where `ERROR_MESSAGE` is available)
- the log file created via `schedule-log`
  -  is locked by resticprofile during the entire process
  - will never be cleaned for new runs of resticprofile

This means, to pipe the log into the Windows event view, I have no other choice but to use the log file, but that log will always contain the content of previous resticprofile runs, making it harder to read with every run.  

I was able to workaround this by adding a second action to the task generated by resticprofile which overwrites the existing log file with an empty one. The action consists of the command: `C:\Windows\System32\cmd.exe` with the parameter: `/C copy NUL  "C:\Path\To\profile-backup.log" /Y`.  

It would be great if this workaround wasn't necessary. I think this could be done by:
- the lock of log file could be freed at some point so that users could make this part of the commands they put into `run-after`
- the restic log could be made available via a variable

jkellerer:
Besides other options (e.g. supporting windows event log), how about:
* On Windows the log file should be opened so that others can still read it (FILE_SHARE_READ or so, needs testing)
* We add a template function `{{ tempfile 'session.log' }}` that returns the path to a temporary file (the same file for the same name within the same run) that can be used in `schedule-log:` and also in `run-after/finally:` and that will be deleted when `resticprofile` ends.

Schroedingers-Cat:
It sounds like all of your suggestions would remove the need for manual changes to the scheduled task, so they all sound good to me.

jkellerer:
@Schroedingers-Cat , would you mind verifying if version 0.21.1 fixes your issue?

Schroedingers-Cat:
@jkellerer thanks for letting me know!

This is how the relevant section of my profile looks:
```toml
  [userdata.backup]
    schedule-log = '{{ .TempDir }}\restic-userdata-backup.log'
    run-after  = "powershell -command Write-MessageEventFromFile -PathToLogFile '$Temp\\restic-userdata-backup.log'"
```

Scheduling that still concatenates the log for each run.  

[...]

PR #174, opened 2023-02-11
Non-locking log appender and template func {{tempFile "filename"}} 

jkellerer:
*PR for*: 
- #168 

**Implements**:
* Log output on Windows uses CRLF
* Non-locking log appender (to support read while writing on a locking FS)
* Log output is flushed just before calling hooks
* Template functions for temp files (and dir) to be used in config files
* Temp files are guaranteed to exist and are writable by the user running resticprofile
* TempDir util func for a temp dir that is auto-removed when resticprofile ends
* Temp dir support for the log flag and the `schedule-log` config

**Example**:

```yaml
global:
  shell: powershell

default:
  repository: "local:test-repo"
  ...
  run-finally:
    - gc "{{ tempFile "rp-run.log" }}" > ".\resticprofile-last.log"
  backup:
    schedule: hourly
    schedule-log: '{{ tempFile "rp-run.log" }}'
```

Use with schedule
```shell
resticprofile schedule backup
```

Test via CLI
```shell
resticprofile --log "temp:/t/rp-run.log" backup
```

### DIFF
--- a/config/profile.go
+++ b/config/profile.go
@@ -2,14 +2,18 @@ package config
 
 import (
 	"fmt"
+	"path"
+	"path/filepath"
 	"reflect"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/creativeprojects/resticprofile/constants"
 	"github.com/creativeprojects/resticprofile/restic"
 	"github.com/creativeprojects/resticprofile/shell"
+	"github.com/creativeprojects/resticprofile/util"
 	"github.com/creativeprojects/resticprofile/util/bools"
 	"github.com/mitchellh/mapstructure"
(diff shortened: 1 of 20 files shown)


creativeprojects:
I fixed the Windows build. I forgot I can't upgrade the `taskmaster` package. I'm supposed to change the code to accommodate the new API but I'm rarely coding on Windows these days...

creativeprojects (approved):
Actually there's no section on logs in the documentation; so we should probably add one at some point. There's no good place for a documentation yet

jkellerer:
That is a good point. Logging might need its own page. But that can come as separate step.

Will also think of improving the documentation of template functions, it is quite basic at the moment, on the other hand template functions in resticprofile are a rather new addition.

PR #180, opened 2023-03-06
Restic: Add restic v15 release

jkellerer:
Restic 15 was released in Jan. JSON schema, reference and CLI args filter is updated with this PR.


jkellerer:
Note: Not much to review. Is the result of adding a new version to the Makefile and committing the build result.

creativeprojects (approved):

PR #181, opened 2023-03-11
Add variables `.OS` and `.Arch` to all templates

jkellerer:
Small PR to add GOOS (`{{ .OS }}`) and GOARCH (`{{ .Arch }}`) to all templates. Also refactored template default data to allow reuse across template data structs.

The primary purpose for this is to allow the creation of reusable configuration, e.g. OS (and/or ARCH) specific `parents`, `mixins` or `includes` where the variables can be used to select what works in a certain context.

### DIFF
--- a/config/template.go
+++ b/config/template.go
@@ -4,25 +4,19 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
-	"strings"
-	"time"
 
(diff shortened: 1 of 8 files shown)


creativeprojects (approved):
It can definitely come handy, I like it 😉 

Thanks 👍🏻

PR #178, opened 2023-02-26
Enhanced: Catch any --help & --dry-run (http only)

jkellerer:
Small followup on #157:

* Consider command args `--dry-run` and `-n` to enable dry-run in http hook
* Catch any `--help` and `-h` arg (no matter at what position it was specified)

E.g. dry-run for http hook is triggered in all of the following cases: 

```shell
# dry-run for resticprofile
resticprofile --dry-run backup

# dry-run for restic
resticprofile backup --dry-run
resticprofile backup -n
```

All of the following cases redirect to resticprofile's help command:

```shell
resticprofile --help backup
resticprofile -h backup
resticprofile help backup
resticprofile backup --help
resticprofile backup -h
```

PR #184, opened 2023-03-15
Build option to disable self update

### LINKED ISSUE #170: Please provide a build time option to disable self-update
for packagers it would be helpful if we could turn off the self update. restic itself solves that via a [tag](https://github.com/restic/restic/blob/master/build.go#L59). maybe a similar solution could be applied here?

creativeprojects:
This is mostly moving code around, except:
- creation of a `OwnCommands` struct so we can register and keep a list of commands
- pass `ownCommands` slice to `Completer` to avoid using a global reference directly (doesn't make a big difference but is more future proof)
- add a `no_self_update` tag to build resticprofile without the self updater

Fixes #170

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,10 @@ build: prepare
 	@echo "[*] $@"
 	$(GOBUILD) -o $(BINARY) -v -ldflags "-X 'main.commit=${BUILD_COMMIT}' -X 'main.date=${BUILD_DATE}' -X 'main.builtBy=make'"
 
+build-no-selfupdate: prepare
+	@echo "[*] $@"
+	$(GOBUILD) -o $(BINARY) -v -tags no_self_update -ldflags "-X 'main.commit=${BUILD_COMMIT}' -X 'main.date=${BUILD_DATE}' -X 'main.builtBy=make'"
+
 build-mac: prepare
 	@echo "[*] $@"
 	GOOS="darwin" GOARCH="amd64" $(GOBUILD) -o $(BINARY_DARWIN) -v -ldflags "-X 'main.commit=${BUILD_COMMIT}' -X 'main.date=${BUILD_DATE}' -X 'main.builtBy=make'"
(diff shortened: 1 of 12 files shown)


jkellerer:
Thanks for requesting a review 😀

Using a custom type for managing the own commands list is in general a good idea!

Only thing I'm not yet sure of is if the help commands fully fits inside of it since it is not only about own commands. Also some changes had been made in "https://github.com/creativeprojects/resticprofile/pull/178/files", e.g. `runOwnCommand` no longer calls `displayHelpCommand` so it is no longer tangled together which I didn't like so much in the first place.

From my opinion, it would be better to keep the help command separate (just like any other command) and only include implementations that relate to command list management (some of the display code may be included though).

creativeprojects:
> Only thing I'm not yet sure of is if the help commands fully fits inside of it since it is not only about own commands. Also some changes had been made in "https://github.com/creativeprojects/resticprofile/pull/178/files", e.g. `runOwnCommand` no longer calls `displayHelpCommand` so it is no longer tangled together which I didn't like so much in the first place.

No problem, I'll wait until you're happy to merge #178 

> From my opinion, it would be better to keep the help command separate (just like any other command) and only include implementations that relate to command list management (some of the display code may be included though).

I totally agree with that: But I've done it like that because we would need to change the signature of the run function to include a reference to the OwnCommands instance. I suppose it kind of make sense but we will only use the reference in the help command.

jkellerer:
Have just merged the PR.

Ah yes, I overlooked that we need this reference. How about changing the signature of `Run` to include a context type. It then is easier to adjust things as the context type can be changed instead?

jkellerer:
What I mean is, change the command func signature to something like `func(io.Writer, CommandRequest) error`

creativeprojects:
Yeah I might as well 👍🏻

creativeprojects:
It certainly looks better 😉

jkellerer (approved):
lgtm!

PR #191, opened 2023-04-05
Updated clog to 0.13

jkellerer:
Also introduced conditional trace generation where useful.

Had verified elevation on Windows, `--trace` still sends trace log from elevated user.


creativeprojects (approved):
LGTM

Nice for the trace on elevated user 😉 

Thanks 👍🏻

Issue #183, opened 2023-03-13
introduce a "base dir" option


JesseRacine:
Currently, resticprofile runs from the current directory that you are in when you run it. I was thinking that there could be a "base-dir" option that would set the directory that resticprofile runs from. If this option were implemented then you could run resticprofile from any directory and the backup would still work, since the backup (source option) could then be relative to the base-dir.

jkellerer:
Makes sense. Could simplify the creation of reusable profiles, when relative paths can be used in source and the base is configured separately.

kohanla:
@jkellerer Thanks for implementing this. Hope it will be merged soon :-). This option is important.

kohanla:
@jkellerer I tested this pull request now. In case anybody else want to test it. Install git and go, then run
```
git clone https://github.com/creativeprojects/resticprofile.git
cd resticprofile
git fetch origin pull/192/head:base-dir
git merge base-dir
make build-all
ls -l resticprofile*
```

It works! Thanks for your code! But I have one wish.

It seems not possible to set the base-dir on "command level" as well. But I would prefer to set the base-dir just for the backup command, because in most cases it is only necessary there.

And then one question. If you do

```
cd /tmp
resticprofile testrestic.restore latest --target .
```

where should resticprofile restore to if base-dir is set to '/'

In my opinion it should be /tmp, but it is for some reason '/'

That's why I would prefer to have an option to set the base-dir on command level. This would allow me to run restore without a base-dir.

jkellerer:
Thanks a lot for testing it. That is very good feedback.

In the current form `base-dir` not only sets the base for source-paths but for anything that depends on the current working dir. For consistency, I think this is correct behaviour, however I also see your use case.

Will update the PR to add a separate option to set a base just for backup sources.

kohanla:
> Will update the PR to add a separate option to set a base just for backup sources.

Thanks! I will test when it is ready.

kohanla:
@jkellerer 
 
> Will update the PR to add a separate option to set a base just for backup sources.

Thanks for your PR! Just to clarify, did you add an option to set a base just for backup sources?

jkellerer:
Yes, there are 2 options, `source-base` and `base-dir`

kohanla:
Thanks! Will test when 0.23 is available.

PR #192, opened 2023-04-06
Allow to set a base-dir inside the profile (#183)

jkellerer:
Implements #183 by changing the current directory when the option is set.

Also updated path documentation (moved description of the configuration path from "configuration/" to "configuration/path" and improved it)

### DIFF
--- a/main.go
+++ b/main.go
@@ -335,6 +335,24 @@ func runProfile(
 		return fmt.Errorf("cannot load profile '%s'", profileName)
 	}
 
+	if len(profile.BaseDir) > 0 {
+		if current, err := os.Getwd(); err == nil {
+			defer func(dir string) {
+				if e := os.Chdir(dir); e != nil {
+					panic(fmt.Errorf("failed restoring working directory %q: %w", dir, e))
+				}
+			}(current)
+		} else {
+			return fmt.Errorf("changing base directory not allowed as current directory is unknown in profile %q; cause: %w", profileName, err)
+		}
+
+		if err = os.Chdir(profile.BaseDir); err == nil {
+			clog.Infof("profile '%s': base directory is %q", profileName, profile.BaseDir)
+		} else {
+			return fmt.Errorf("cannot change to base directory %q in profile %q; cause: %w", profile.BaseDir, profileName, err)
+		}
+	}
+
 	displayProfileDeprecationNotices(profile)
 	c.DisplayConfigurationIssues()
 
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"math/rand"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"runtime"
 	"syscall"
 	"text/tabwriter"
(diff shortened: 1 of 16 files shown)


jkellerer:
Needs a separate base just for backup-sources. See #183

jkellerer:
@creativeprojects , this PR is ready for review now. Let me know what you think.

creativeprojects:
It looks good to me 👍🏻 

I have a question though. It might be a bit far fetched, but what about a configuration like this:

```yaml
profile:
    backup:
        source-base: "/tmp"
        source: "{{ .CurrentDir }}"
```

`CurrenDir` as well as `$PWD` refers to the _original_ current dir.
Would it make sense to change the values from `source-base` as well? I'm not sure

creativeprojects:
Actually I realise `base-dir` is changing `$PWD` but not `.CurrentDir`. We might want to do both for consistency?

jkellerer:
Think it depends how `.CurrentDir` is interpreted. If we define it as the directory where `resticprofile` was started, then it shouldn't be changed.

jkellerer:
... if we define it as the directory that runs shell hooks, we should change it.

jkellerer:
Thinking about it again.. the cleanest way is likely:
* `.CurrentDir` always points to the current working dir... this is `base-dir` if set and like now when unset.
* `.StartupDir` (or something similar) is the directory where `resticprofile` was started. This is the same as `.CurrentDir` unless `base-dir` is set.

creativeprojects:
Agree.

I know this could be a little confusing that `.CurrentDir` will have different values during the lifecycle of the configuration, but I suppose it would be easy to explain why in the documentation.

jkellerer:
Will update the PR later today

creativeprojects:
Excellent, thanks 👍🏻

jkellerer:
Thanks for taking the time of reviewing it!

main.go
```diff
@@ -315,6 +316,48 @@ func setPriority(nice int, class string) error {
...
+		}
+
+		done = true
+
+		// Adjust baseDir if needed
+		if len(profile.BaseDir) > 0 {
+			var currentDir string
+			currentDir, err = os.Getwd()
+			if err != nil {
+				err = fmt.Errorf("changing base directory not allowed as current directory is unknown in profile %q: %w", profileName, err)
```
creativeprojects, line 337: Is this a problem to change the current directory to `BaseDir` then leave without changing it back to the (unknown) original?

I'm not sure there are valid cases where it could happen though. The only one I can think of is executing a shell script while downloading it (via a unix pipe). I suppose why not?

main.go
```diff
@@ -315,6 +316,48 @@ func setPriority(nice int, class string) error {
 	return nil
 }
 
+func openProfile(c *config.Config, profileName string) (profile *config.Profile, cleanup func(), err error) {
+	cleanup = func() {}
+	done := false
+
+	for attempts := 3; attempts > 0 && !done; attempts-- {
```
creativeprojects, line 323: why 3 attempts?

main.go
creativeprojects: come to think of it... you still get a current directory
```
/tmp % echo "pwd" | sh
/tmp
```

jkellerer:
@creativeprojects do you think it is ok, or do we need changes 😄 ? Thanks in advance.

creativeprojects:
Sorry I forgot about this PR,
Yes this is good, thank you 👍🏻

creativeprojects (approved):

PR #193, opened 2023-04-10
New option to specify Prometheus Pushgateway job name

spasche:
Hi,

First of all, thanks for the project, it is very helpful!

I tried the Prometheus Pushgateway feature for monitoring and noticed an issue in case you work with multiple profiles.
When running multiple backups in a row, only the last backup metrics are returned from the Pushgateway.
That's the behavior of the Pushgateway to group all metrics by the job name specified in the URL (or any other labels in the URL, details are on https://github.com/prometheus/pushgateway#api).

To prevent such issues, I propose adding an option to specify the name of the Pushgateway job (currently, this is always set to the command, e.g. `backup`). This way, it is possible to prevent metrics from profile A interfering with metrics from profile B.

Example configuration:
```yaml
version: "2"

mixins:
  prom:
    prometheus-push: http://prometheus-pushgateway.monitoring:9091/
    prometheus-push-job: "${COMMAND}-{{ .Profile.Name }}"

profiles:
  default:
    repository: /backups/root
    use:
      - prom
    backup:
      extended-status: true
      source:
        - /

  documents:
    repository: /backups/documents
    use:
      - prom
    backup:
      extended-status: true
      source:
        - /user/documents
```

If you're good with the approach, I can also update the documentation.

Thanks!

### DIFF
--- a/monitor/prom/progress.go
+++ b/monitor/prom/progress.go
@@ -1,6 +1,10 @@
 package prom
 
 import (
+	"fmt"
+	"os"
+	"strings"
+
 	"github.com/creativeprojects/clog"
 	"github.com/creativeprojects/resticprofile/config"
 	"github.com/creativeprojects/resticprofile/constants"
@@ -55,7 +59,17 @@ func (p *Progress) Summary(command string, summary monitor.Summary, stderr strin
 		}
 	}
 	if p.profile.PrometheusPush != "" {
-		err := p.metrics.Push(p.profile.PrometheusPush, command)
+		jobName := p.profile.PrometheusPushJob
+		if jobName == "" {
+			jobName = fmt.Sprintf("%s.%s", p.profile.Name, command)
+		}
+		jobName = os.Expand(jobName, func(name string) string {
+			if strings.EqualFold(name, "command") {
+				return command
+			}
+			return ""
+		})
+		err := p.metrics.Push(p.profile.PrometheusPush, jobName)
 		if err != nil {
 			// not important enough to throw an error here
 			clog.Warningf("pushing prometheus metrics to %q: %v", p.profile.PrometheusPush, err)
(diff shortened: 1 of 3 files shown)


jkellerer:
Thanks a lot for this contribution, that is well spotted.

monitor/prom/progress.go
```diff
@@ -55,7 +57,12 @@ func (p *Progress) Summary(command string, summary monitor.Summary, stderr strin
 		}
 	}
 	if p.profile.PrometheusPush != "" {
-		err := p.metrics.Push(p.profile.PrometheusPush, command)
+		jobName := "${COMMAND}"
+		if p.profile.PrometheusPushJob != "" {
+			jobName = p.profile.PrometheusPushJob
+		}
+		jobName = strings.ReplaceAll(jobName, "${COMMAND}", command)
```
creativeprojects, line 64: It looks good.
But for maximum compatibility I think we can also use `os.Expand()` instead of `strings.ReplaceAll()`.
So we can also use $COMMAND as not everybody likes to enclose environment variable around curly braces

creativeprojects:
Yes it's good for me as well 👍🏻 

> If you're good with the approach, I can also update the documentation.

Yes please 😉 

Thanks

jkellerer (changes requested):

monitor/prom/progress.go
jkellerer: Had actually started a review with exactly the same comment :) .. but forgot to commit it...

jkellerer:
Thanks for trying. Not using my PG either for backup so far and haven't verified it. 
When you query the PGs pull address do you see metrics of both profiles at the same time?

This case was about that when running profiles in a group (that run faster than the pull interval of prometheus), only the last profile will be visible.

creativeprojects:
Yeah, actually it looks like push gateway is only keeping the last label in memory, which is weird.

When I said I could query both set of metrics, I meant that Prometheus had enough time to load both into its own storage.

jkellerer:
@spasche please check the review from above. It is just a small update regarding default value and the way how variables are resolved.

monitor/prom/progress.go
```diff
@@ -55,7 +57,12 @@ func (p *Progress) Summary(command string, summary monitor.Summary, stderr strin
...
+		err := p.metrics.Push(p.profile.PrometheusPush, jobName)
```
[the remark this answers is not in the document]
spasche: I agree it's better to have a default value which works ouf of the box. Suggestion applied, thanks!

jkellerer (approved):

PR #195, opened 2023-04-17
Global json schema (auto versioning)

jkellerer:
This PR:
* Generates a JSON schema that redirects to versioned schemas based on `version` and `global.restic-version` properties (see also [SchemaStore](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md)). 
* Adds an URL for the schema: https://creativeprojects.github.io/resticprofile/jsonschema/config.json
* Fixes `generate --config-reference filename`
* Simplifies config-1 schema files (schema in `additionalProperties` instead of via regex matching with negative lookbehind)

The is the first step to add the schema to the schemastore catalog (we need a single schema for this) and is also more convenient if the URL doesn't need to be changed 😀

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ RESTIC_GEN=$(BUILD)restic-generator
 RESTIC_DIR=$(BUILD)restic-
 RESTIC_CMD=$(BUILD)restic-commands.json
 
+CONTRIB_DIR=contrib
 JSONSCHEMA_DIR=docs/static/jsonschema
 CONFIG_REFERENCE_DIR=docs/content/configuration/reference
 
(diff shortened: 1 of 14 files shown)


jkellerer:
After some adjustments it works well for YAML and JSON now. "Even better TOML" doesn't support `if/then/else`, but without it, YAML and JSON report errors incorrectly, since not setting a version is also an error when only using [implication](https://json-schema.org/understanding-json-schema/reference/conditionals.html#implication) and it is the first and only reported error in this case.

One improvement for "Even better TOML" however is that the updated version-1 schema now also works correctly. Previously it didn't support profile commands due to the lack of full regex support in the editor plugin.

For testing with local references, run:

````
SCHEMA_BASE_URL="http://localhost:1313/resticprofile/jsonschema" \
resticprofile generate --config-reference contrib/templates/config-schema.gojson > docs/static/jsonschema/config.json
````

Auto selection then works with:

````yaml
# yaml-language-server: $schema=http://localhost:1313/resticprofile/jsonschema/config.json
version: "2"

global:
  restic-version: "0.15"

````

creativeprojects:
This looks good to me, thanks for this PR 👍🏻 

I was trying the new configuration locally but it seems that it doesn't recognise the version `2`. Am I doing something wrong?

Screenshot 2023-04-22 at 23 13 53

jkellerer:
Might be due to implication. While most examples indicate that it should be fine to use it for versioning, I think it needs to be updated to select all versions first using `if/then/else` and only when this validates, choose a schema for the rest of the document. Otherwise editors will always struggle with priority of errors (or fail to select a different schema).

Will update the template.

creativeprojects:
Thanks!

I had a quick look at the JSON schema specs and hmmm it does look cutting edge 😐

creativeprojects:
MR too old to rebase, replaced by #412 
It does work on `yaml` files on VSCode now, thanks for the work 👍🏻

PR #197, opened 2023-04-19
Templates: Add `map`/`splitR`/`contains`/`matches`

jkellerer:
Adds a few more template functions:

* `{{ if "some string" | contains "some" }} some is there {{ end }}`
* `{{ if "some string" | matches "some|thing" }} some or thing is there {{ end }}`
* `{{ $token := range ( "x ; y , z" | splitR "\\s*[,;]\\s*" ) }} {{ $token }} {{ end }}`

Mapping variant 1: 

````
{{ define "tpl" }}
Hello {{ .whom }}
{{ end }}
{{ template "tpl" ("whom" "World" | map) }}
````

Mapping variant 2:

````
{{ define "tpl" }}
Hello {{ ._0 }}
{{ end }}
{{ template "tpl" (list "World" | map) }}.
````

### DIFF
--- a/docs/content/configuration/templates/index.md
+++ b/docs/content/configuration/templates/index.md
@@ -579,9 +579,12 @@ Please note there are some functions only made available by hugo though, resticp
 ## Template functions
 
 resticprofile supports the following set of own functions in all templates:
-
+* `{{ "some string" | contains "some" }}` => `true`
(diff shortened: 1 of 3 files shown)


creativeprojects (approved):
LGTM 😉 
Thanks 👍🏻

Issue #153, opened 2022-11-18
Add 'stdout-command' to 'dump' command


ozboss:
Please add a 'stdout-command' (analogous to 'stdin-command' flag of the backup command) to the 'dump' command. Together with #151 this would allow to restore for example some application database by simply calling `resticprofile application_db.dump latest` to restore the latest backup the application database.

jkellerer:
Thats an interesting idea, though 'stdin-command' is primarily provided to allow scheduled backup of streams. Nevertheless I agree that it is more convenient when also the restore can be done with a single command call where only the snapshot ID (e.g. latest) has to be provided.

ozboss:
Yes exactly, this would allow to restore the stream similarly to how it was backup up. My backup in this case is a mysqldump. Restic provides a good example for this exact [backup](https://restic.readthedocs.io/en/stable/040_backup.html#reading-data-from-stdin) and [restore](https://restic.readthedocs.io/en/stable/050_restore.html#printing-files-to-stdout) process.

jkellerer:
I'll add `stdout-command` and `stdout-file` to all restic commands, so that their output can be redirected cleanly. This can not only benefit `dump` but also other restic commands that may generate JSON output. 

We may also add `stdout-is-binary: (true|false)` to control whether output is logged or printed when the `--quiet` flag isn't set (binary output makes no sense when it isn't clean, printing it may in text mode just cause failures). This would then default to `true` for `dump` (or any other command that generates binary data).

damoclark:
> We may also add stdout-is-binary: (true|false) to control whether output is logged or printed when the --quiet flag isn't set (binary output makes no sense when it isn't clean, printing it along with other output may just cause failures). This would then default to true for dump (or any other command that generates binary data).

I just got tricked up with this, attempting to restore a filesystem image snapshot using resticprofile with dump command.  Was baffled when it overflowed the device file saying no space left on device.  Only when I inspected the output of the dump that I twigged that resticprofile output was at the start.  

D.

PR #207, opened 2023-05-01
Ft redirect output

jkellerer:
Work in progress:
- [ ] Implement #153 for all commands (provide clean stdout redirection)
- [ ] Auto enable "extended-status" if required or warn if status/prom feature is non-functional
- [ ] Provide a console monitor (when "extended-status" output is enabled)
- [ ] Update restic output parsers to cover current versions
- [ ] Eval if `prune`, `check`, `forget` can be covered in output parsing


creativeprojects:
> Provide a console monitor (when "extended-status" output is enabled)

I've been thinking about doing that for quite some time as well 😉

jkellerer:
Sorry didn't wan't to take it away, but it just fit too well into rethinking output handling.

creativeprojects:
> Sorry didn't wan't to take it away, but it just fit too well into rethinking output handling.
> While the status monitor already works quite well, the output needs to be improved.

No problem 👍🏻 
FYI I was thinking either copying the restic code used to display the progress, or go wild with a library like https://github.com/pterm/pterm

jkellerer:
... I was thinking I'd like to learn how it can be achieved, so I created it mostly from scratch. No external deps but maybe less smart.

256 colors could be supported and there's also an option to read the background on xterm to decide on the theme. Eventually this can be useful but for the moment I try to limit the amount of colors that are used. In the end it shouldn't be annoying.

creativeprojects:
> 256 colors could be supported and there's also an option to read the background on xterm to decide on the theme. Eventually this can be useful but for the moment I try to limit the amount of colors that are used. In the end it shouldn't be annoying.

I agree. Too many colours is complicated to get right on both a light and dark background anyway.

Issue #211, opened 2023-05-23
Issue with systemd


d11ahirst:
Hi,

I am having problems with the scheduling systemd timer,service

This is my profiles.toml

```
[default]
  initialize = true
  repository = "{{ .Env.RESTIC_REPOSITORY }}"
  password = "{{ .Env.RESTIC_PASSWORD }}"
  option = "s3.connections=3"

  [default.env]
    AWS_ACCESS_KEY_ID = "revoked"
    AWS_SECRET_ACCESS_KEY = "revoked"

  [default.backup]
    exclude = [ ...... ]
    exclude-caches = true
    one-file-system = true
    tag = [ "root" ]
    source = [ "/foldertobackup" ]
    schedule = ["hourly","daily"]
    schedule-permission = "system"
    schedule-priority = "background"
    schedule-lock-wait = "10m"
    schedule-log = "backup.log"

  [default.restore]
     target = "/"

  [default.retention]
    before-backup = false
    after-backup = true
    keep-hourly = 12
    keep-daily = 7
    keep-weekly = 1
    keep-tag = [ "manual", "locked" ]
```

I get the following output in the log
   
```
2023/05/23 09:50:18 INFO  profile 'default': initializing repository (if not existing)
reading repository password from stdin
2023/05/23 09:50:18 INFO  profile 'default': starting 'backup'
Fatal: unable to open config file: stat <no value>/config: no such file or directory
Is there a repository at the following location?
<no value>
2023/05/23 09:50:18 ERROR backup on profile 'default': exit status 1
```

Any help figuring this out would be appreciated.

d11ahirst:
Thanks for the reply,

What you suggested worked fine being able to read the env variables in schedules what be great.

jkellerer:
Variable capture for schedules is coming in #212

northben:
I just encountered this issue as well. I'm on version 0.24.0 and I can't find any way to get the scheduled backup job to read environment variables.

PR #212, opened 2023-05-23
Schedule: Capture `os.Environ` on schedule creation

jkellerer:
Fixes: #211 

The PR adds a new config option `schedule-capture-environment` defaulting to `RESTIC_*` which enables OS environment capturing for schedules at creation time.

@creativeprojects - what do you think (it is an enhancement to the existing capture of the profile env)? 
Does this work with #146 ?

### DIFF
--- a/config/profile.go
+++ b/config/profile.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"fmt"
+	"os"
 	"path"
 	"path/filepath"
 	"reflect"
@@ -10,6 +11,7 @@ import (
 	"time"
 
 	"github.com/Masterminds/semver/v3"
+	"github.com/creativeprojects/clog"
 	"github.com/creativeprojects/resticprofile/constants"
 	"github.com/creativeprojects/resticprofile/restic"
 	"github.com/creativeprojects/resticprofile/shell"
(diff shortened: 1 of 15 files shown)


jkellerer:
@creativeprojects , btw. not all platforms have case insensitive variables. But this PR makes env vars case insensitive in all cases. Should we differentiate it by platform? 

(I think yes we should)

creativeprojects:
> what do you think (it is an enhancement to the existing capture of the profile env)? Does this work with #146 ?

I don't think it would clash with #146 so we're good 👍🏻 

> not all platforms have case insensitive variables. But this PR makes env vars case insensitive in all cases. Should we differentiate it by platform?

Yeah probably 😐  I'm not too familiar with powershell but I'm pretty sure that could cause some issues.

jkellerer:
Env vars will now preserve the case. Case folding only applies on Windows. Will manually test the changes first.

jkellerer:
Works fine in launchd. Systemd test is pending

jkellerer:
Think it can be merged. @creativeprojects , what do you think? 

Please check the environment variables part. I think it works correct now, given the limitation that the config doesn't allow to specify the case, but all unix OS are case sensitive for env vars. 

The implementation was adjusted to this:
* If the config defines a variable that exists in the environment, the config value overrides the existing var when the name matches. Case is not considered, there is a resolve step for this that adjusted the name.
* If the config defines a new variable, it will be all upper-cased like it was before.

creativeprojects (approved):
Yes it works well, nice one!

I think we're good to go with this one as well 👍🏻 

Thank you for the great work 😉

Issue #215, opened 2023-05-26
Can't escape `E:\$RECYCLE.BIN` exclude on Windows


Ambient-Impact:
Hi there. First off, thank you for this project because it seems to be really well thought out and incredibly useful, and I know how much work FOSS stuff can be.

I'm running into an issue on Windows where I'm trying to back up an entire local drive (`E:\`) which is just lots of files, so not operating system or running programs. I'd prefer to pass restic the drive root so that I don't have to remember to add any new folders in the drive to my back ups and just exclude a couple of operating system created directories to avoid errors. The one that's giving me problems is the recycle bin directory, named `E:\$RECYCLE.BIN` which I can't seem to escape or quote correctly in PowerShell no matter what I try. The backtick (`` ` ``) seems to have no effect. My workaround for now is just to manually list all the top level folders I want to include in the back up, but other wrappers such as [prestic](https://github.com/ducalex/prestic) seemed to handle this fine and I kind of miss that.

Current work in progress messy `profile.yaml`:

```yaml
# @see https://creativeprojects.github.io/resticprofile/configuration/examples/#configuration-example-for-windows

version: '1'

global:
#   shell: 'bash'
  priority: 'idle'
  restic-arguments-filter: false

default:
  repository: 'rclone:lan-server:backup_desktop/restic'
  password-file: 'C:\Users\i\restic\restic-lan-server-password.txt'
  initialize: false

  backup:

    verbose: true

    source:
      - E:\
      # - 'E:\Design'
      # - 'E:\Disk Images'
      # - 'E:\Files'
      # - 'E:\MonitoRSS back ups'
      # - 'E:\Repository'
      # - 'E:\Web'

    exclude:
      - 'E:\$RECYCLE.BIN'
    #   # - '"E:\$RECYCLE.BIN"'
      - 'E:\System Volume Information'

    # # files with no path are relative to the configuration file
    # exclude-file:
    #   - exclude.txt

    check-after: true

    # run-before: dir /l
    # run-after: echo All Done!

    # ignore restic warnings
    # without it the backup is considered failed when restic can't read some files
    no-error-on-warning: true

    # Every day at 11 AM.
    #
    # @todo Can this be set to run after the Acronis back up has completed
    #  via a post command in the Acronis options?
    schedule: '*-*-* 11:00:00'

    # @todo Use VSS snapshot; requires backup permissions.
    # @see https://restic.readthedocs.io/en/stable/040_backup.html
    #use-fs-snapshot: true

[...]

PR #216, opened 2023-05-28
Variables: Allow to escape "$" with "$$"

jkellerer:
Addresses: #215 

Note: There is no documentation section on variables in values using "${var}" syntax (besides a note for mixins). It might make sense to add such a section to [variables](https://creativeprojects.github.io/resticprofile/configuration/variables/index.html).


jkellerer:
@creativeprojects , let me know what you think regarding documentation. I can enhance the PR to document `${var}` usage.

creativeprojects (approved):
Yes, I like it 😉 
We can definitely add some documentation on how `${var}` are resolved 👍🏻

PR #220, opened 2023-06-08
Fix 194: Allow to set "keep-tag", “tag” & “group-by” as empty string

### LINKED ISSUE #194: [Question] How to configure forget with empty keep-tag
Usually i do not tag my backups.
So tags on snapshots have an special meaning and should be keep forever.
Restic implents this with an empty `--keep-tag ''` (see the info section under [remove snapshots docu](https://restic.readthedocs.io/en/stable/060_forget.html#removing-snapshots-according-to-a-policy)),
this also applies to the `--tag` paramter (further down in the docu).

I can not get this to work with resticprofile.
My naive approach was to use `keep-tag: "''"` but this ends in generating `--keep-tag=\'\'` and restic shows `Applying Policy: ... and all snapshots with tags [['']]` and doesn't work.

So maybe someone could help me with this?

jkellerer:
Fixes #194 & #232

Example:

```yaml
  my-profile:
    inherit: default
    backup:
      source:
        - /path/to/backup
    retention:
      after-backup: true
      keep-last: 3
      keep-tag: ""
```

Results in:
```
➜ resticprofile --dry-run my-profile.backup
...
dry-run: /usr/local/bin/restic forget --compression=auto --keep-last=3 --keep-tag="" ...
```

Also supports `--tag=""` and `--group-by=""` which also have a special meaning for retention.

### DIFF
--- a/config/flag.go
+++ b/config/flag.go
@@ -22,7 +22,9 @@ func init() {
 }
 
 var allowedEmptyValueArgs = []string{
-	constants.ParameterKeepTag,
+	constants.ParameterKeepTag, // allows --keep-tag="" - means keep all with any assigned tag
+	constants.ParameterTag,     // allows --tag=""      - means match all untagged snapshots
+	constants.ParameterGroupBy, // allows --group-by="" - means do not group snapshots
 }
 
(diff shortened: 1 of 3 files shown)


creativeprojects (approved):
Looks good to me 👍🏻  sorry I'm very late reviewing

PR #221, opened 2023-06-08
Fix: Do not add `--tag` for `tag: true`

jkellerer:
Fixes a problem, found while testing #220.

`tag` is implicitly true in `retention` from config version 2 and the following config:

```yaml
version: "2"
profiles:
  my-profile:
    inherit: default
    backup:
      source:
        - /path/to/backup
    retention:
      after-backup: true
      keep-last: 3
```

...results in:
```
➜ resticprofile --dry-run my-profile.backup
...
dry-run: /usr/local/bin/restic forget ... --repo="..." --tag
```

But for `tag` (as well as `path` and `host`), `true` means copy from backup, not set an empty flag.

With the pr, no empty `--tag` flags are added for `tag` being boolean `true`.


creativeprojects (approved):
Also good to go 😄 
Thanks for the fix
(sorry it took me so long...)

PR #222, opened 2023-06-08
Fix 218: Args filter must not remove paths

### LINKED ISSUE #218: The first path from source gets filtered out on backup
Hello. I've run into an issue when the first path in the config is ignored when I create backups. Here's my minimal config to reproduce this bug:

```toml
version = "1"

[default]
password-command = 'echo password'
initialize = true
no-error-on-warning = true
repository = 'backup'

[default.backup]
source = [
    'test-folder',
    'test-folder-2'
]
```

Note the line `no-error-on-warning = true`. When I remove this line, everything works as expected.

## Steps to reproduce
1. Use the above config
2. Create 2 folders and name them "test-folder" and "test-folder-2"
3. Put some files into these folders
4. Run the command `resticprofile -v default.backup` and see that the "test-folder" is present in the "unfiltered command:" line, but is absent in the "starting command:" line:
image
5. Run the command `resticprofile default.snapshots` and see that the snapshot is created without the "test-folder" path:
image

## Workarounds
- Don't use the `no-error-on-warning = true` option in the config
- Add the `-v` key to the end of the backup command. For example `resticprofile default.backup` then becomes `resticprofile default.backup -v`
- Add a dummy path to the config like this:
```toml
[default.backup]
source = [
    '', # dummy
    'test-folder',
    'test-folder-2'
]
```

## Versions
- restic 0.15.2 compiled with go1.20.3 on windows/amd64
- resticprofile version 0.22.0 commit ddf9debf89960db2c6139523ef6140a6538c5d27

jkellerer:
Fixes #218

### DIFF
--- a/wrapper.go
+++ b/wrapper.go
@@ -308,7 +308,7 @@ func (r *resticWrapper) validArgumentsFilter(validArgs []string) argumentsFilter
 					if !includesValue {
 						skipValue = false
 					}
-				} else {
+				} else if !allowExtraValues { // if "allowExtraValues" => args with values must use "--arg=value"
 					skipValue = !includesValue
 				}
 				continue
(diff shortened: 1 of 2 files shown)


creativeprojects (approved):
That was definitely a weird bug 😆 
Good job finding the issue 👍🏻

PR #224, opened 2023-06-10
Fix 223: Escape absolute restic path for pwsh

### LINKED ISSUE #223: Cannot use `powershell` as shell when resticprofile is placed in a path containing whitespaces (Windows)
I tried replacing a `powershell --command` call in my profile's `run-after` setting by adding `powershell` as `shell` value to the global section. But jobs scheduled like that fail to run with the following error:
```
+ c:\program files\tools\restic.exe backup --exclude-caches --exclude-f ...
+ ~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (c:\program:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

I have resticprofile in a subdirectory of `C:\Program Files`, which contains a whitespace and there seems to be an issue with the way the full path is placed into the task scheduler. I tried wrapping the path to resticprofile with `"` but that didn't help.

jkellerer:
Fixes #223

Also limits use of autoenv feature for pwsh to existing env vars and introduces `$RESTICPROFILE_PWSH_NO_AUTOENV` to disable it completely

### DIFF
--- a/shell/command.go
+++ b/shell/command.go
@@ -14,6 +14,8 @@ import (
 
 	"github.com/creativeprojects/clog"
 	"github.com/creativeprojects/resticprofile/monitor"
+	"github.com/creativeprojects/resticprofile/util/collect"
+	"golang.org/x/exp/slices"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


jkellerer:
This PR needs an enhancement, see #230

jkellerer:
1ef2c01035ed7666438aff68cb98cdebe2b9e098 also fixes #230

creativeprojects:
It looks good to me, thank you 😉 

I guess we need a little documentation about the new environment variable 👍🏻 

BTW I was testing it on a new Win 11 machine and I realise git bash no longer works, or maybe it has never worked properly. This is out of the scope of this PR though, but we should look at it at some point.

```
2023/07/16 21:36:39 unfiltered command: backup --exclude=*.exe --exclude=.git --exclude=examples\private --password-file=examples\key --repo=r:\backup --verbose=1 .
2023/07/16 21:36:39 starting command: C:\Users\fred\go\bin\restic.exe backup --exclude=*.exe --exclude=.git --exclude=examples\private --password-file=examples\key --repo=r:\backup --verbose=1 .
/usr/bin/bash: C:Usersfredgobinrestic.exe: command not found
2023/07/16 21:36:39 backup on profile 'self': exit status 127
exit status 1
```

creativeprojects (approved):

jkellerer:
For bash, "C:\Users\fred\go\bin\restic.exe" is tricky since backslash is escape we cannot easily fix it. We might fix it like it was done for PWSH when `command` points to an executable file but still other parameters would potentially suffer from the same issue. 

Windows and bash/linux mixing support might need its own case (if required). On the other hand, with WSL2 in Win11 I'm not sure how relevant this is. WSL2 works quite ok for the linux part.

Regarding documentation: I think we're missing a page for shells and the difference in behaviour in general. For the moment, I'd keep it a hidden tuning variable.

Issue #225, opened 2023-06-10
Configuring a retention on a stdin profile applies the retention policy to the whole repository


Lucki:
I'm new to restic and maybe I'm interpreting the command output wrong. 
I have a few profiles in the configuration. Some normal path profiles and a stdin profile. One profile has no retention configured.

The retention output for the normal profile looked correct, listing all snapshots of that profile only and declaring the reasons to keep each one.
However, the output of the forget command from a stdin profile got me thinking:
`[…] keep 2 snapshots: […] Reasons: daily within 7d […]` is listed for a third profile, although that specific profile has no retention. In fact that command listed all profiles with the same retention policy.

The difference seems to be the `--path=…` command line flag which is missing for the profile using stdin.

Retention command after normal profile:
~~~ sh
starting command: /usr/bin/restic forget […] --path=/home/lucki --prune --repo=local:/run/media/lucki/backup/restic --verbose=1
~~~
Retention command after stdin profile which applies to all profiles including a profile without retention:
~~~ sh
starting command: /usr/bin/restic forget […] --prune --repo=local:/run/media/lucki/backup/restic --verbose=1
~~~

My uneducated guess is that the absence of the `--path=` flag will apply the retention policy to the whole repository regardless of other configured retention policies.

Reference example: https://creativeprojects.github.io/resticprofile/configuration/examples/index.html#use-stdin-in-configuration

<details>
<summary>My trimmed config</summary>

~~~ toml
version = "1"

[default]
    repository = "local:/run/media/lucki/backup/restic"
    verbose = true

[home]
    inherit = "default"

    [home.backup]
        source = [
            "/home/lucki",
        ]

    [home.retention]
        before-backup = false
        after-backup = true
        keep-within-yearly = "5y"
        keep-within-monthly = "1y"
        keep-within-weekly = "1m"
        keep-within-daily = "7d"
        prune = true

[gitea]
    inherit = "default"

    [gitea.backup]
        stdin-filename = "gitea.sql"
        stdin-command = [
            "pg_dump …",
        ]

    [gitea.retention]
        before-backup = false
        after-backup = true
        keep-within-yearly = "5y"
        keep-within-monthly = "1y"
        keep-within-weekly = "1m"
        keep-within-daily = "7d"
        prune = true

[third]
    inherit = "default"

    [third.backup]
        source = [
            "/mnt/third",
        ]

        # Notice *no* retention!
~~~

</details>

----
Bonus question:
[...]

PR #227, opened 2023-06-12
Retention: Align host filter with "backup"

jkellerer:
Aligns the `--host` filter of "retention" (forget) with "backup" in restic (if host is not set, it defaults to [`os.Hostname()`](https://github.com/restic/restic/blob/237f32c651d4924242783a02c9a52b09ccc2dc4f/cmd/restic/cmd_backup.go#L48) in restic, but forget has no default matching all hosts if unset)

After this change, when using `retention` in version 2 configs, all 3 filters `path`, `tags` and `host` are aligned with the `backup` section if not redefined.

Other changes:
* Auto enables `after-backup: true` if nothing was specified explicitly but `keep-*` is set
  (to avoid case like #226).
* Fixes a problem with the `tag` copy from `backup`. Tags must be copied as single flag (`"tag1,tag2,tag3,..."`) instead of multiple flags to match correctly
  (mentioned in #225)

@creativeprojects : before I create tests, what do you think on this?

### DIFF
--- a/config/info_customizer.go
+++ b/config/info_customizer.go
@@ -72,18 +72,31 @@ func init() {
 			info.mayBool = true
 			info.examples = []string{"true", "false", fmt.Sprintf(`"%s"`, propertyName)}
 
-			if sectionName == constants.CommandBackup {
-				if propertyName != constants.ParameterHost {
-					info.examples = info.examples[1:] // remove "true" from examples of backup section
-					note = fmt.Sprintf(`Boolean true is unsupported in section "backup".`)
-				}
-			} else {
-				note = fmt.Sprintf(`Boolean true is replaced with the %ss from section "backup".`, propertyName)
-			}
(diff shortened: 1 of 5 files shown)


creativeprojects:
>  before I create tests, what do you think on this?

Yes this is a good idea 👍🏻 

I was also thinking we should fix the inconsistency of `retention.after-backup` and `backup.check-after` but that's out of the scope of this MR.

Sorry I'm very busy this week, and the next one as well. We're deploying a new project to production next week 😉

jkellerer:
@creativeprojects , think the PR is ready for review now. What do you think?

creativeprojects (approved):
LGTM 😄 

BTW we should keep track of all the subtle differences between v1 and v2 so we can make a exhaustive upgrade documentation

Thanks 😉

jkellerer:
Guess we’ll need to review usages of `config.Version` when we document differences. This time it is in the reference at least 🙂

Issue #233, opened 2023-07-26
Compile instructions are missing dependencies, at least eget and mockery.


AphidGit:
On a freshly installed linux system, running; 

`zypper install -y go git make makeinfo autoconf automake gcc gcc-c++`

Then trying to compile by hand: 

```
git clone https://github.com/creativeprojects/resticprofile.git 
cd resticprofile 
make build 
```
results in: 
```
generate.go:4: running "eget": exec: "eget": executable file not found in $PATH
config/info.go:20: running "mockery": exec: "mockery": executable file not found in $PATH
```
The instructions at https://creativeprojects.github.io/resticprofile/installation/source/index.html are out of date, there appear to be some undocumented dependencies. 

Edit: These commands appear to do the trick: 

``` 
go install github.com/zyedidia/eget@latest
PATH=$PATH:~/go/bin
```

jkellerer:
@AphidGit , would you mind trying https://github.com/jkellerer/resticprofile/tree/fix-mockery-warning (PR #234)?

PR #234, opened 2023-07-30
Support build when GOPATH is unset / fix mockery build warning

jkellerer:
Changes

* Extends the documentation on how to build (includes info that `eget` and `mockery` are installed)
* Use more absolute paths (do not depend on GOBIN being in PATH)
* Apply suggested mockery config to silence deprecation warning
* Add ARM64 for all 3 OS (for builds with `make`, goreleaser already had it)
* Might also fix: #233

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,13 @@ ifeq ($(GOBIN),)
 endif
 
 BINARY=resticprofile
-BINARY_DARWIN=$(BINARY)_darwin
-BINARY_LINUX=$(BINARY)_linux
+BINARY_DARWIN_AMD64=$(BINARY)_darwin
+BINARY_DARWIN_ARM64=$(BINARY)_darwin_arm64
+BINARY_LINUX_AMD64=$(BINARY)_linux
+BINARY_LINUX_ARM64=$(BINARY)_linux_arm64
 BINARY_PI=$(BINARY)_pi
-BINARY_WINDOWS=$(BINARY).exe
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -56,29 +59,41 @@ TOC_END=<\!--te-->
...
+	$(GOCMD) install -v github.com/zyedidia/eget@latest
 
-$(GOBIN)/goreleaser: $(GOBIN)/eget
+$(GOBIN)/goreleaser: verify $(GOBIN)/eget
 	@echo "[*] $@"
-	eget goreleaser/goreleaser --upgrade-only --to $(GOBIN)
+	"$(GOBIN)/eget" goreleaser/goreleaser --upgrade-only --to $(GOBIN)
 
-$(GOBIN)/github-markdown-toc.go: $(GOBIN)/eget
+$(GOBIN)/github-markdown-toc.go: verify $(GOBIN)/eget
```
creativeprojects, line 83: this was quite handy for a while but I don't think we need this tool anymore 😉

creativeprojects (approved):
Nice one! I was going to do it this week but you were faster 😆 
Thanks again

Makefile
jkellerer: Didn’t check for orphan steps though in this PR.

creativeprojects: Don't worry about it, I'll clean it up some time

jkellerer:
Merging it now. If #233 is still an issue we can open another PR to refine.

PR #235, opened 2023-07-31
Detect if running on battery

creativeprojects:
Add a flag to avoid starting a profile when the system is running on battery.
Works for all supported OS.

Adds 2 new options in the configuration:
* `schedule-ignore-on-battery` boolean (default false)
* `schedule-ignore-on-battery-less-than` int => battery charge left

Adds a new command line flag `--ignore-on-battery`:
* if no parameter, it just stops when the battery is on discharge
* with a parameter, only when the charge is lower or equal than the value (and discharging)

### DIFF
--- a/battery.go
+++ b/battery.go
@@ -1,0 +1,85 @@
+package main
+
+import (
+	"math"
+
+	"github.com/creativeprojects/clog"
+	"github.com/distatus/battery"
+)
+
+// IsRunningOnBattery returns true if the computer is running on battery,
+// followed by the percentage of battery remaining. If no battery is present
+// the percentage will be 0.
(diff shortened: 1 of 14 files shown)


jkellerer:
Would it make sense to allow setting a SOC? 
E.g. `--ignore-on-battery` or `--ignore-on-battery 75` (if SOC equals to or is lower than 75%)?

creativeprojects:
> Would it make sense to allow setting a SOC?

Excellent idea 👍🏻 

BTW the command line flag is added to be compatible with `v1` configuration otherwise we wouldn't really need it. Although I suppose some user might still prefer using their own scheduling way...

PR #238, opened 2023-08-03
Restic: Add restic v16 release

jamebus:
Hello! Restic 0.16.0 has been released, here is updated command line options for that.


jkellerer:
@jamebus, thanks for your contribution and no worries about the test failures. I have fixed them in another branch (PR #239) that is already merged, just rebase and we should be good to go.

jamebus:
@jkellerer thanks! I was fixing the tests, but you totally beat me to it. Your solution is better anyhow, i wasn't using regexp to filter out the unwanted commands, making it a bit unnecessarily complicated.

Issue #242, opened 2023-08-08
iexclude-file only has relative path on command line


jdwhite:
`resticprofile version 0.23.0-dev commit 6cc332d362bec58570a6730a70b54ed7d360b1dc`

When using `iexclude-file` only the relative path is being passed to restic and restic will fail to find it. This is unlike `exclude-file` which gets the absolute path.
Example using both exclude-file and iexclude-file:

```
starting command: /usr/bin/restic backup --exclude-file=/root/.config/resticprofile/excludes.global --iexclude-file=iexcludes.global <...rest removed for brevity...>
open repository
repository 4fe32096 opened (version 1)
lock repository
open iexcludes.global: no such file or directory
```

Workaround is to specify the absolute path to the `iexclude-file` in the configuration.

creativeprojects:
Oops that's a miss 😨 

Well spotted 👍🏻

jkellerer:
@creativeprojects , do you want to fix it or should I. Just checked there are 2 more to add (arrived in restic 0.10 and 0.12).

creativeprojects:
I don't mind either way, it looks simple enough to fix. What are the other ones?

creativeprojects:
excellent, thank you 👍🏻

PR #243, opened 2023-08-08
Fix 242: iexclude-file not converted to abs path

jkellerer:
fix for #242 , also fixes `files-from-raw` & `files-from-verbatim`


jkellerer:
Case came at the right time :)

PR #249, opened 2023-08-15
Allow to interrupt the wait for a lock

creativeprojects:
In the current implementation it is not possible to interrupt resticprofile when it's waiting for a lock (both kinds of locks).

This small PR returns an `interrupted` error when the wait is cancelled (typically by a `CTRL`-`c`)

- [x] Implementation
- [x] Tests
- [x] Ready for review

### DIFF
--- a/command_error.go
+++ b/command_error.go
@@ -24,6 +24,10 @@ func (c *commandError) Error() string {
 	return c.err.Error()
 }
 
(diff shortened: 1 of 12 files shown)


creativeprojects:
We would get a better implementation by passing a `context` around but that's a **lot** of changes 😐

PR #251, opened 2023-08-19
Reduce log level for `Getwd()` failures (fixes #250)

### LINKED ISSUE #250: Bug: failed retrieving pwd: %sstat .: permission denied
When running any `resticprofile` command as a user that does not have access to the current working directory, you get the following error:

```
❯ sudo -u restic resticprofile -h |head -6   # inside my normal user
failed retrieving pwd: %sstat .: permission denied

resticprofile is a configuration profiles manager for backup profiles and is the missing link between a configuration file and restic backup

Usage:
  resticprofile [resticprofile flags] [profile name.]restic-command [restic flags]
```

This is a problem for me because I run my all my restic commands with another user "restic". This breaks the output for example like this:
```
❯ sudo -u restic resticprofile -q -n local snapshots --json |jq
parse error: Invalid literal at line 1, column 7
```

jkellerer:
Fix for #250

Reduce log level for `Getwd()` failures in template data
Also delay logging in init to ensue logger is setup

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/adrg/xdg v0.4.0
 	github.com/capnspacehook/taskmaster v0.0.0-20210519235353-1629df7c85e9
 	github.com/coreos/go-systemd/v22 v22.5.0
-	github.com/creativeprojects/clog v0.13.0
+	github.com/creativeprojects/clog v0.14.0
 	github.com/creativeprojects/go-selfupdate v1.1.1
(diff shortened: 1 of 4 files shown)


jkellerer:
Waiting on https://github.com/creativeprojects/clog/pull/12

creativeprojects:
We could maybe test with a virtual filesystem but that's getting a bit complicated
Thanks 👍🏻

PR #268, opened 2023-10-15
Docs for installation from scoop

hgraeber:
resticprofile is available for installation from the [scoop](https://scoop.sh) main bucket. Automatic update for new versions is supported as long as download and hash paths do not change except for the version number.


creativeprojects:
Excellent! thank you for adding resticprofile to the scoop repo 👍🏻

PR #270, opened 2023-10-21
systemd: prevent paging in systemd schedules

jkellerer:
Small PR to add `--no-pager` to all `systemctl` calls.

In most systemd based Linux, `systemctl`  will start a pager after every created schedule when running `resticprofile schedule --all` (or when multiple schedules are in a profile). 

With many schedules this is not very user friendly, especially since `systemctl` output is not very long and doesn't need a pager.


creativeprojects:
Nice catch! 👍🏻 thanks

creativeprojects (approved):

PR #276, opened 2023-10-29
Profile: support source with "-" (dash) prefix

### LINKED ISSUE #264: Relative source path with glob starting with dash doesn't work
Let have two profiles:
* 1st excluding paths starting with `-`,
* 2nd include only them.

Simplified sample (without `inherit`, `tags`, etc…), can be like:
```yaml
profiles:
  one:
    backup:
      exclude: '-*'
      source: '.'

  two:
    backup:
      source:
        - '-*'     # ignored
        - './-*'   # ignored
        - './-withoutGlob'  # only this source works
```

The problem is that with the following conditions:
1. relative path
2. uses glob
3. starting with dash

`resticprofile` passes empty sources to `restic` (checked with `--dry-run` option).
So users have to put explicit paths (`./-withoutGlob`), which makes config less maintainable.

jkellerer:
Fixes #264

### DIFF
--- a/config/profile.go
+++ b/config/profile.go
@@ -633,9 +633,17 @@ func (p *Profile) resolveSourcePath(sourceBase string, sourcePaths ...string) []
 		applyBaseDir = absolutePrefix(p.BaseDir)
 	}
 
+	// prefix paths starting with "-" with a "./" to distinguish a source path from a flag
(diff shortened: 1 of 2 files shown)


creativeprojects (approved):
Well spotted! thanks 👍🏻

PR #277, opened 2023-10-29
Logging: allow to setup default log output in global

### LINKED ISSUE #261: Allow to set resticprofile `--log` in config file
There is `schedule-log` to redirect logs for scheduled tasks, but for manual run users have to use `--log` flag. This makes backup command longer than it can be. Also in this case some options are placed in config file, but some (`--log`) should be added by hand to `resticprofile` call. Is there a reason for this?

In my case some automated backups are scheduled and `schedule-log` works perfect, but some other depend on an event (some job finished and it's time to backup) and I have to run `resticprofile` with `--log` flag to keep things unified. If logs path changes, this change must be updated in two different places, which isn't good.
It would be great to set logging the same way for both cases (or at least put both options inside one config).
Is it possible to add `log` option to config?

jkellerer:
Fixes #261

Adds a new option to define the default log output in global:

```yaml
global:
  log: 'some-destination'
```

Adds a new destination syntax `-` for the case that:
* global redirects to a file
* you want to redirect to console or stdout instead (`-` means "console or stdout")

### DIFF
--- a/commands_display.go
+++ b/commands_display.go
@@ -158,15 +158,6 @@ func displayResticHelp(output io.Writer, configuration *config.Config, flags com
 	out, closer := displayWriter(output, flags)
 	defer closer()
 
-	// try to load the config
-	if configuration == nil {
-		if file, err := filesearch.FindConfigurationFile(flags.config); err == nil {
-			if configuration, err = config.LoadFile(file, flags.format); err != nil {
-				configuration = nil
-			}
-		}
-	}
-
 	resticBinary := ""
 	if configuration != nil {
 		if section, err := configuration.GetGlobalSection(); err == nil {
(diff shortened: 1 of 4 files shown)


creativeprojects (approved):
It looks good to me, thanks 👍🏻 

I tried windows elevation and the logs are going through fine 😛 

That was also one of the last bits missing for the `run-schedule` command, cool 😎

creativeprojects:
I was thinking... at some point we might want to distinguish regular commands and interactive commands like `self-update` because it doesn't make much sense to send the output somewhere else than the console 😐

jkellerer:
Yes. Though these commands might still use log as is but output the important information to stdout / console instead of using log (which is mostly the case already, might just need a review)

PR #281, opened 2023-11-01
Add `prometheus-push-format` to allow selecting text format

watsonbox:
The default golang client push format [is `FmtProtoDelim`](https://github.com/prometheus/client_golang/blob/f030c3d062e5180830979221733d46d286790bac/prometheus/push/push.go#L112C15-L112C35).

However, some implementations do not support protocol buffers, [including VictoriaMetrics](https://github.com/VictoriaMetrics/VictoriaMetrics/issues/3745).

This PR adds the ability to select a simple text format in such cases.

By default (protocol buffers):

image

With `prometheus-push-format: text`:

image

### DIFF
--- a/config/profile.go
+++ b/config/profile.go
@@ -94,6 +94,7 @@ type Profile struct {
 	PrometheusSaveToFile    string                            `mapstructure:"prometheus-save-to-file" description:"Path to the prometheus metrics file to update with a summary of the last restic command result"`
 	PrometheusPush          string                            `mapstructure:"prometheus-push" format:"uri" description:"URL of the prometheus push gateway to send the summary of the last restic command result to"`
 	PrometheusPushJob       string                            `mapstructure:"prometheus-push-job" description:"Prometheus push gateway job name. $command placeholder is replaced with restic command"`
+	PrometheusPushFormat    string                            `mapstructure:"prometheus-push-format" default:"protobuf" enum:"protobuf;text" description:"Prometheus push gateway request format"`
 	PrometheusLabels        map[string]string                 `mapstructure:"prometheus-labels" description:"Additional prometheus labels to set"`
 	SystemdDropInFiles      []string                          `mapstructure:"systemd-drop-in-files" default:"" description:"Files containing systemd drop-in (override) files - see https://creativeprojects.github.io/resticprofile/schedules/systemd/"`
 	Environment             map[string]ConfidentialValue      `mapstructure:"env" description:"Additional environment variables to set in any child process"`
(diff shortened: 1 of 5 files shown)


jkellerer:
Thanks a lot, wasn’t even aware that it uses protobuf by default. 

Should we make “text” the default format? 

The efficiency gains of the binary protocol shouldn’t be too much in this use case so the default could be a setting that works in all cases.

watsonbox:
> Should we make “text” the default format?

It did cross my mind, but I figured you'd be less keen on that due to it being a (potentially) breaking change. In reality it seems likely that endpoints accepting protocol buffers will also accept plain text, but I've really no idea.

watsonbox:
I suppose from a consistency POV it also makes sense that the default format used in this repo matches the default format used by the Prometheus client library.

jkellerer:
Ok, agree. Thanks a lot!

creativeprojects (changes requested):
Agree it makes sense to keep the library default.

It looks good to me,
But could we add a little note in the documentation about it?

watsonbox:
> But could we add a little note in the documentation about it?

Done, although in attempting to do so I actually came across [this Prometheus documentation](https://prometheus.io/docs/instrumenting/exposition_formats/#exposition-formats) which I hadn't seen before. I'm not sure if it changes the reasoning above, but it does make a slightly more compelling case for simply changing this to text.

creativeprojects:
> NOTE: Some earlier versions of Prometheus supported an exposition format based on [Protocol Buffers](https://developers.google.com/protocol-buffers/) (aka Protobuf) in addition to the current text-based format. As of version 2.0, however, Prometheus no longer supports the Protobuf-based format. You can read about the reasoning behind this change in [this document](https://github.com/OpenObservability/OpenMetrics/blob/main/legacy/markdown/protobuf_vs_text.md). However, beginning with Prometheus v2.40, there is experimental support for native histograms, which – at least in its initial experimental state – utilizes the old Protobuf format (with some newer additions) again. Therefore, a very recent Prometheus server with the experimental native histogram support enabled, will again be able to ingest the Protobuf format. This support is experimental and might get removed again.

I'm tempted to use `text` as default then.

As a breaking change, if any, users would discover very quickly that the metrics are no longer going into their monitoring system, and would easily change it back to `protobuf`.
No data were killed 😆 

Thoughts?

watsonbox:
Makes sense 👍. I made the update.

creativeprojects:
Thanks!

@jkellerer are you OK with changing the default to `text`?

creativeprojects (approved):

jkellerer:
@creativeprojects , yes `text` is fine, was also my original thought. Since the extra info seems to suggest that protobuf is more a legacy or exceptional format, using `text` is the way to go and prevents that this config has to be adjusted (in most cases) which should always be the goal 😀. 

Still funny that the default client uses it by default which was a valid argument for keeping it but as long a more compatible option is available we should use this (as default).

PR #283, opened 2023-11-01
Remove rclone binary in make clean target

watsonbox:
Looks like it may have been overlooked in https://github.com/creativeprojects/resticprofile/commit/35d07ee34f7f0175e38ac9c9dce688738f521694.


creativeprojects:
It was definitely missed, thanks for this PR 👍🏻

Issue #288, opened 2023-11-11
Support supercronic as scheduler


meyayl:
I am trying to run Resticprofile on Openshift. 

Our compliance (enforced by kyverno) requires us to run payloads with unprivileged user permissions and no privilege escalation, which seem to make it impossible to run `crond` inside an alpine container. The same is true for systemd.

[supercronic](https://github.com/aptible/supercronic) on the other hand runs cronjobs for the current user (including unprivileged users), its file format is basically the know cron format.

Please add support for supercronic, so that users that find themselves in a situation where they need to meet specific compliance challenges can use this wonderful tool as well :)

jkellerer:
As far as I understand [supercronic](https://github.com/aptible/supercronic) needs a crontab file.

When you configure resticprofile with:

```yaml
global:
  scheduler: crond
```

It expects an executable to manage the crontab file, its interface is:

```
crontab -l > crontab.txt
cat new-crontab.txt | crontab -
```

jkellerer:
We could also add support to specify a crontab file since all is already in place we just use the `crontab` binary to select the correct file depending on the context.

meyayl:
Relying on crontab might provide two challenges: 
1. I am not sure if setting the suid flag on the `crontab` binary will run on the OpenShift with our restricted policies.
2. We are mandated to use read-only fil esystems, so we would need to be able to set the target path.

A solution where we could freely specify the target location for the generated contab file, would allow storing the file on an ephemeral volume. 

If I am not mistaken, this would allow generating the contab file content with `resticprofile -c  /path/to/profile schedule`  and then `exec supercronic /path/to/generated/crontab`  to hand over to supercronic and benefit from scheduled job execution.

meyayl:
I should have read you first comment completely. 

The wrapper replaces the crontab binary, so I don't need to set a suid flag and won't run into problems with the read-only file system. 

I can work with that :)

I still feel it might be nice to have direct support for supercronic on the long run, so that people can benefit from it out of the box.

Thank you for your quick response!

creativeprojects:
> We could also add support to specify a crontab file since all is already in place we just use the crontab binary to select the correct file depending on the context.

The `crontab` binary is also sending a signal to `crond` to reload the configuration files. If we save a crontab directly we'll need to do it ourselves 😉

jkellerer:
Btw. I'm updating the `crond` support:
[...]

PR #289, opened 2023-11-11
crond: add support for crontab file only (on any OS)

jkellerer:
Implements #288

The PR allows to use an external scheduler that supports crontab format and adds the following configuration options:

```yaml
global:
  # writes schedules to "file" and detects if a user column is needed
  # detection scan the file contents (header & entries, default to no)
  scheduler: "crontab: /path/to/crontab/file"

  # crontab with user column using user "root"
  scheduler: "crontab:root:/etc/cron.d/resticprofile"

  # crontab with user column using the current user
  scheduler: "crontab:*:/etc/cron.d/resticprofile"

  # crontab without user column
  scheduler: "crontab:-:/var/spool/cron/crontabs/user"

  # using the specified binary to read and save the crontab
  scheduler: "crond:/usr/bin/crontab"
```

(this is available on any OS)

The existing `crond` support using the `crontab` binary remains the same but it is now used as fallback when available (`crontab` binary is found) and `systemd` is not available.

~waiting on #259 before implementing remaining tests~


creativeprojects:
This would also help with #321 as I realised that the crond implementation in `busybox` (alpine linux) has funny rules when trying to run as a non-root user. It doesn't work with the current implementation of using `crontab` to set or update the scheduling.

jkellerer:
Had it tested all options manually, think it is ready now.

docs/content/schedules/_index.md
```diff
@@ -10,16 +10,21 @@ resticprofile is capable of managing scheduled backups for you. Under the hood i
...
 
 {{< tabs groupid="config-with-json" >}}
 {{% tab title="toml" %}}
 
 ```toml
 [global]
   scheduler = "crond"
+  # scheduler = "crond:/usr/bin/crontab"
+  # scheduler = "crontab:*:/etc/cron.d/resticprofile"
+  # scheduler = "crontab:-:/var/spool/cron/crontabs/username"
```
creativeprojects, line 27: we should probably have a small section explaining the parameters, specially the middle one 🤔

jkellerer: Or a link to the reference / global section as it is mentioned there. However was also thinking it may be needed.

creativeprojects: A link to the reference would probably do 😄

jkellerer: added both 🙂

creativeprojects (approved):
LGTM

I'll try to use it with a non-root user container and see if that helps 👍🏻 

Thanks 😉 

-- Edit
I re-read the original post that triggered this PR... that's exactly what the OP wanted to do via Supercronic 🤦🏻 🙄

docs/content/schedules/_index.md
```diff
@@ -54,7 +65,7 @@ global:
 {{% /tab %}}
 {{< /tabs >}}
 
-
+See also [reference / global section]({{% relref "/configuration/reference/#section-profilebackup" %}}) for options on how to configure the scheduler.
```
creativeprojects, line 68:
```suggestion
See also [reference / global section]({{% relref "/configuration/reference#section-profilebackup" %}}) for options on how to configure the scheduler.
```

Sorry the links no longer work with a trailing `/` 🤷🏻

jkellerer:
Yes. Supersonic is one option to have a compatible task scheduler that is OS independent (besides that it needs to run as service)

jkellerer:
The use case with alpine should work as well. Had used busybox’s cron with custom crontab files previously.

creativeprojects:
> The use case with alpine should work as well. Had used busybox’s cron with custom crontab files previously.

Yes it works when running as root.
With this PR I managed to get resticprofile running as a non-privileged user (using Supercronic).
I'll add a contrib for it.

I'm also tempted to provide another docker image with running resticprofile as a non-privileged user.

Issue #282, opened 2023-11-01
[feature request] Allow profiles and sections to be disabled


snstamml:
As resticprofile execute a restic command "blindly", accidently running an unconfigured command like "forget" could be fatal. As a safety measure, it would be helpful if a section could be disabled deliberately, e.g.:

```
myprofile:
  forget:
    enabled: false
```

Similarly, as a convenience, a complete profile could be disabled in the same way:

```
myprofile:
  enabled: false
```

jkellerer:
Was also thinking about something like this as there are use cases where a dedicated profile may be created just for a subset of commands. 

Just not yet sure about how to configure it most efficiently. Disabling individual sections is one option but if you want to enforce that only a handful of commands is used it wouldn’t scale very well. 

Maybe an include / exclude list of available commands might be better e.g. that could also be used to prevent sections of a base (e.g. default) profile from getting used accidentally.

creativeprojects:
Funny enough this week I was considering introducing dot profiles but then I remembered it might not work with TOML.

When you need a base profile that you can't really run because it's incomplete, you prefix it with a dot:

```yaml

.base_profile:
  source: /

real_profile:
  inherit:
    - .base_profile
```

- you cannot run any command on a dot profile
- you don't see them in `show` command 

Need to find a nice way to do that in TOML though 🤔

jkellerer:
Maybe use a prefix like “__” (e.g. __default) and allow to configure the base profile prefix in global for the case that anyone needs to use this name. (Would have also been the case for dot prefixes)

A limit of available commands would also be helpful in addition but the base profile filter by prefix would simply disable all commands (using the name is a nice solution as it doesn’t inherit and this can even be used within the inheritance chain)

creativeprojects:
Making it configuration with a default of `__` would certainly work 👍🏻  I like that

jkellerer:
Ok in this case we have 2 different ways of configuring it:

The original proposal:

```yaml
profiles:
  default:
    enabled: false

  check-only:
    inherit: default
    enabled: true    # must be set as default is disabled
    backup:
      enabled: false

  backup-only:
    inherit: default

    prune:
      enabled: false
    restore:
      enabled: false
    mount:
      enabled: false

   my-backup:
     inherit: backup-only
     enabled: true    # must be set as default is disabled
```

Is straight forward but very verbose. 

Was thinking about something like this as an alternative (@creativeprojects, @snstamml what do you think?):

```yaml
[...]

PR #290, opened 2023-11-12
base-profiles: supports profiles with limited/no command usage

jkellerer:
Implements #282 

~Waiting on #259 before completing it~

Related updates in command completion will come separately

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -22,6 +22,7 @@ import (
 	"github.com/creativeprojects/resticprofile/restic"
 	"github.com/creativeprojects/resticprofile/schedule"
 	"github.com/creativeprojects/resticprofile/term"
+	"github.com/creativeprojects/resticprofile/util"
 	"github.com/creativeprojects/resticprofile/util/templates"
 	"github.com/creativeprojects/resticprofile/win"
 )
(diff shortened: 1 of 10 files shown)


jkellerer:
@creativeprojects, the PR is updated to include command filter (& base-profile) support in `runSchedule` and I've extended it to allow running V2 schedules. It still misses more tests and a manual test.

creativeprojects:
> Note: I've removed the comment that V2 configs will disallow inline schedules since I think this limitation isn't required. After all inline schedules are the only option to inherit configs and base-profiles make it possible to do so without side effects.

hmmm true; do we still need a separate schedule section after all? because we could allow the same schedule section to groups, as this was the missing piece basically.

creativeprojects:
or maybe just moving all `schedule-*` options into a `scheduling` subsection, attached to a profile or a group. Maybe that would be nicer after all

jkellerer:
Well I think a separate schedule section does make sense as the runtime behaviour differs. Inline schedules all run individually, but the schedule section runs as a single item. No sure if that can be fully implemented with groups.

jkellerer:
@creativeprojects , had just drafted how a unified `Schedule` instance could look like. For `toml` the differences are quite little. Think I changed my mind and like option 2 more as it fits more natural in the current config file format. Nevertheless regardless of which may we may go, it shouldn't matter for this PR. In any case the code added here can be adjusted and reused. Let us continue discussion in #80.

PR #319, opened 2024-02-13
Add prometheus label to build info

creativeprojects:
Add the profile label, and any other label from the configuration to the `resticprofile_build_info` metric

Fixes: https://github.com/creativeprojects/resticprofile/discussions/318


Javex:
Amazing! You're quick! I just rolled it out on one server and ran it. No more log errors from `node_exporter`! I will keep it on just one server and monitor this PR for any issues that might come up until a new version is released.

Thanks for such a quick fix, that's a massive help!

Issue #316, opened 2024-02-12
Environment variables from file


mjmeintjes:
I'd like to get the sensitive environment variables (like S3 keys) from a file instead of having them directly in the config file.

So similar to the `password-file` option, maybe have an `env-file` in addition to the current `env` option.

creativeprojects:
Yeah I suppose that shouldn't be difficult to implement 🤔

jkellerer:
@creativeprojects , I'll take this one.

jkellerer:
... and for later maybe: re-evaluate profile if env changed.

PR #323, opened 2024-02-18
env-file: Implement dotenv and {{env}} support

jkellerer:
Implementation of #316

Supporting:

* `env-file` in profiles:

   ```yaml
     my-profile:
       env-file: /path/to/.env
   ```

* `{{env}}` function to get a temporary .env file that can be used to adjust the environment between shell command runs:

   ```yaml
     my-profile:
       run-before: echo "VAR=WORLD" >> "{{env}}"
       backup:
         run-before: echo "Hello $VAR"
   ```


creativeprojects:
I can see there are still some broken links in the documentation since the latest version of Hugo 😞 
Maybe I should just pin the version for now 🤔 that's annoying

creativeprojects (commented):
GitHub doesn't seem to let me add comments on lines that haven't been changed, or am I going mad? 🤔  

Anyway... in the functions `runShellCommands` and `runFinalShellCommands` we create the environment only once for the list of commands. So if we declare a variable in the first command, we wouldn't have it available in the second one (and so forth).

I suppose it could be useful in case we want to do something like:

```yaml
run-before:
  - "echo \"USER=GUEST\" >> {{ env }}"
  - "echo \"HOME=/home/$USER\" >> {{ env }}"
```

...which is actually a terrible example because `godotenv` library would take care of it for us 🤣 

Do we need to restore the environment on each line?

jkellerer:
Rebased and updated so that env is created for every shell command invocation. Done from my side.

creativeprojects (approved):
LGTM.
Tried on my side and it works all fine 👍🏻 

Although I was thinking about moving the "setting up environment variables" debug to trace level instead as it's becoming quite verbose 🤔

jkellerer:
Agree that it all became too verbose with the `--verbose` flag enabled. Still I think changes to the env is an important log output. I've reduced unfiltered-log to trace level and changed the env log to display the full env only when it was changed.

E.g. it now looks like this:

```
/resticprofile --verbose containers.backup
2024/03/10 14:09:29 resticprofile 0.27.0-dev
2024/03/10 14:09:29 using configuration file: /etc/resticprofile/profiles.yaml
...
2024/03/10 14:09:29 profile 'containers': initializing repository (if not existing)
2024/03/10 14:09:29 command environment:
TMPDIR=/tmp

2024/03/10 14:09:29 starting command: /usr/bin/restic init ...
2024/03/10 14:09:29 starting run-before backup on profile 1/1
2024/03/10 14:09:29 command environment: reusing previous
Create a readonly snapshot of '/opt/containers' in '/opt/snapshot_containers'
2024/03/10 14:09:29 profile 'containers': starting 'backup'
2024/03/10 14:09:29 command environment: reusing previous
2024/03/10 14:09:29 starting command: /usr/bin/restic backup ... /opt/snapshot_containers
```

creativeprojects:
> Agree that it all became too verbose with the `--verbose` flag enabled. Still I think changes to the env is an important log output. I've reduced unfiltered-log to trace level and changed the env log to display the full env only when it was changed.

Perfect!
Thanks 😉

PR #336, opened 2024-03-09
restic: use repository-file if the repo flag contains a password

jkellerer:
Experimental PR to replace `restic --repo https://user:password@host` with `restic --repository-file /tmp/profile-repo.txt` (with file mode `0600`)

Currently fails tests (as they don't expect replacement) and depends on some functions from other PRs to complete it.

### DIFF
--- a/config/confidential_test.go
+++ b/config/confidential_test.go
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/creativeprojects/resticprofile/util/templates"
+	"github.com/stretchr/testify/require"
+	"os"
 	"reflect"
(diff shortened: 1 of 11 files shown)


creativeprojects:
I guess this is to hide the password if someone looks at the process list?

Makes sense 👍🏻

jkellerer:
Exactly. A process list may also be sent by monitoring software so its easy to leak passwords around.

jkellerer:
Tests are fixed and the enhancement disables itself when it detects that the repo is set by env variables

creativeprojects (approved):
I tried it on one of my REST repo and it worked like a charm 🥇 

Nice one! thanks 😉

Issue #229, opened 2023-06-18
Nice in systemd unit is always 5


jee-r:
Hello 
I noticed that `nice` in global section is not correctly [passed](https://github.com/creativeprojects/resticprofile/blob/3476fbce1425f3368527c05a9130d511fe530139/systemd/generate.go#L31) to systemd unit template. 

for example with this config

```yaml
version: "1"
global:
  nice: 17
default:
  repository: "my_repo"
  password-file: "key"
  backup:
    exclude-file: "excludes"
    exclude-caches: true
    one-file-system: true
    tag:
      - "test"
    source:
      - "/tmp/"
    schedule: "*-*-* 12:15:00"
    schedule-permission: system
    schedule-priority: background
    schedule-lock-mode: default
    schedule-lock-wait: 15m30s
 ```
After running `schedule` the systemd unit file is 

```
cat /etc/systemd/system/resticprofile-backup@profile-default.service
[Unit]
Description=resticprofile backup for profile default in ./profiles.yaml

[Service]
Type=notify
WorkingDirectory=/tmp
ExecStart=/usr/local/bin/resticprofile --no-prio --no-ansi --config ./profiles.yaml --name default --lock-wait 15m30s backup
Nice=5
Environment="HOME=/root"
Environment="SUDO_USER=test"
```
Nice is set to 5 and not to 17 as expected

**Comment by creativeprojects** [comment]
This is right, I can see in `systemd` that nice is set to a predefined value:
https://github.com/creativeprojects/resticprofile/blob/3476fbce1425f3368527c05a9130d511fe530139/systemd/generate.go#L121

Thanks for the report 👍🏻

**Comment by flolilo** [comment]
Hello,
Just wanted to chime in that when starting resticprofile from the terminal with:

```
# leaving out definitely unrelated settings, i.e. paths
global:
  # neither this:
  priority: "idle"
  # nor this:
  nice: 19
  # ...nor both together change anything.

default:
  backup:
    verbose: 2
```
also will create restic instances with the default niceness.

- `resticprofile version`: resticprofile version 0.24.0 commit bcfaaa7cacfa7f88a94d5dc62ca69b2d89205d89
- `restic version`: restic 0.16.2 compiled with go1.21.3 on linux/amd64

scy:
> will create restic instances with the default niceness

Can confirm, I'm seeing this too.

* resticprofile version 0.26.0 commit 2ab1f3a38467114aeb4095c0924817b369280e39
[...]

PR #339, opened 2024-03-13
Fix group priority

creativeprojects:
- Fix a bug when setting the process priority on Linux, macOS and unixes in general
- Refactoring to avoid duplicate code
- should also work inside a container

### DIFF
--- a/priority/linux.go
+++ b/priority/linux.go
@@ -53,13 +53,13 @@ func SetNice(priority int) error {
 	// group variants of Setpriority etc to affect all of our threads in one go
 	err = unix.Setpgid(pid, 0)
 	if err != nil {
-		return fmt.Errorf("cannot set process group priority, restic will run with the default priority: %w", err)
+		return fmt.Errorf("cannot set process group, restic will run with the default priority: %w", err)
 	}
 
 	clog.Debugf("setting process priority to %d", priority)
-	err = unix.Setpriority(unix.PRIO_PROCESS, pid, priority)
+	err = unix.Setpriority(unix.PRIO_PGRP, pid, priority)
 	if err != nil {
-		return fmt.Errorf("cannot set process priority, restic will run with the default priority: %w", err)
+		return fmt.Errorf("cannot set group process priority, restic will run with the default priority: %w", err)
 	}
 	return nil
 }
@@ -69,9 +69,18 @@ func SetClass(class int) error {
 	return SetNice(class)
 }
 
-// GetNice returns the current nice value
-func GetNice() (int, error) {
+// GetProcessNice returns the nice value of the current process
+func GetProcessNice() (int, error) {
 	pri, err := unix.Getpriority(unix.PRIO_PROCESS, 0)
+	if err != nil {
+		return 0, err
+	}
+	return 20 - pri, nil
+}
+
+// GetProcessNice returns the nice value of the current process group
+func GetGroupNice() (int, error) {
+	pri, err := unix.Getpriority(unix.PRIO_PGRP, 0)
 	if err != nil {
 		return 0, err
 	}
--- a/priority/linux.go
+++ b/priority/linux.go
@@ -51,7 +51,7 @@ func SetNice(priority int) error {
 		return fmt.Errorf("cannot set process group, restic will run with the default priority: %w", err)
 	}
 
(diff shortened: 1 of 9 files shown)


priority/linux.go
```diff
@@ -34,32 +34,31 @@ const (
...
-	err = unix.Setpriority(unix.PRIO_PROCESS, pid, priority)
+	clog.Debugf("setting group process priority to %d", priority)
+	err = unix.Setpriority(unix.PRIO_PGRP, selfPID, priority)
 	if err != nil {
-		return fmt.Errorf("cannot set process priority, restic will run with the default priority: %w", err)
+		clog.Debugf("setting process priority to %d instead", priority)
+		err = unix.Setpriority(unix.PRIO_PROCESS, selfPID, priority)
+		if err != nil {
+			return fmt.Errorf("cannot set process priority, restic will run with the default priority: %w", err)
+		}
```
scy, line 61: I know that your intent is

> Try to be more resilient in case of a failure

but does this really help? As a user, I'm setting the niceness in order to not have the restic run impact other processes on the system too much. But the thing that's using the resources is restic, not resticprofile. Changing just resticprofile's priority doesn't solve the problem, in my opinion.

If you ask me, resticprofile should not consider "well, at least I could set _my own_ priority" a success and instead flat out stop running if my desired configuration could not be applied. Because otherwise, people won't notice. I mean, as far as I can tell, the broken implementation has been in resticprofile for four years and people didn't notice ;)

(Also, you're no longer logging the original `err` that prevented the process group `Setpriority` call.)

priority/unix.go
```diff
@@ -10,20 +11,31 @@ import (
...
+	}
+
+	clog.Debugf("setting group process priority to %d", priority)
+	err = unix.Setpriority(unix.PRIO_PGRP, selfPID, priority)
+	if err != nil {
+		clog.Debugf("setting process priority to %d instead", priority)
+		err = unix.Setpriority(unix.PRIO_PROCESS, selfPID, priority)
+		if err != nil {
+			return fmt.Errorf("cannot set process priority, restic will run with the default priority: %w", err)
+		}
```
scy, line 38: Same remarks as above.

scy (commented):
Thanks for working on this!

I currently don't have a working Go setup to test this, sorry. Instead I've done a quick review of the proposed changes and hope that I'm helping by providing some feedback.

priority/linux.go
scy: Hm, makes sense. But doesn't https://github.com/creativeprojects/resticprofile/issues/229#issuecomment-1859349273 (i.e. the report of restic _not_ running with the configured niceness) show that the priority is not being inherited, at least in some cases?

creativeprojects: That's right, but I don't really understand why 🤔

priority/unix.go
```diff
@@ -10,20 +11,31 @@ import (
...
-	err = unix.Setpriority(unix.PRIO_PROCESS, pid, priority)
+	// Move ourselves to a new process group so that we can use the process
+	// group variants of Setpriority to affect all of our processes at once
+	err = unix.Setpgid(selfPID, 0)
 	if err != nil {
-		return fmt.Errorf("error setting process priority: %v", err)
+		return fmt.Errorf("cannot set process group, restic will run with the default priority: %w", err)
+	}
+
+	clog.Debugf("setting group process priority to %d", priority)
```
scy, line 31: Note that this also says "group process" and is not fixed by bf118a0887f158b1126ec385a298fac4db2c4d92.

Sorry, I thought that "same remarks as above" in my other comment would make it clear that _both_ my remarks (the typo and the fallback behavior) apply; that was apparently not clear enough, my bad.

priority/linux.go
scy: Me neither, sorry :man_shrugging:

I suppose the code as you're proposing it fixes the `PRIO_PGRP` issue and the fallback doesn't make things _worse_ compared to before, so … good enough for me :wink:

creativeprojects:
- [x] refactoring to remove duplicates
- [x] cleanup tests
- [x] run tests inside a goroutine (it doesn't seem to make any difference)
- [x] use default process group
- [x] creates own process group only if current group fails: should fix priority inside containers
- [x] verified restic is running with expected priority

Issue #217, opened 2023-05-29
[Feature Request] Allow systemd randomize RandomizedDelaySec parameter in the schedule command


darkblaze69:
Allow systemd randomize RandomizedDelaySec parameter in the schedule command.

This is a convenient way to avoid starting the jobs on all hosts at one time if config is the same for them with the same schedule time.

creativeprojects:
Hi,

I don't think I would want to backup multiple machines at the same time (to the same repository).
In case of different repositories it doesn't matter if they start at the same time.

I mean I don't mind adding this feature but I'm not sure it's advised to do so

creativeprojects:
closing as it can be set via a systemd template: https://creativeprojects.github.io/resticprofile/schedules/systemd/index.html#how-to-change-the-default-systemd-unit-and-timer-file-using-a-template

scy:
Hi! I've stumbled across the same issue just now, trying to add some randomness to the schedule.

> I don't think I would want to backup multiple machines at the same time (to the same repository).

Which is one of the reasons why [`RandomizedDelaySec`](https://www.freedesktop.org/software/systemd/man/latest/systemd.timer.html#RandomizedDelaySec=) makes sense. It basically allows you to use the same config on multiple machines, but have systemd spread out the iterations over some configurable interval.

For example, you could have 10 machines configured to run a daily backup at 2am, but add `RandomizedDelaySec=2h` to have these backups _actually_ start between 2am and 4am.

> In case of different repositories it doesn't matter if they start at the same time.

That's not quite true. Imagine having 10 machines at home that start uploading their backups at the same time, although to different off-site repos. You'd probably run into congestion of your upstream link. This, too, could be solved by adding a random delay.

> closing as it can be set via a systemd template: https://creativeprojects.github.io/resticprofile/schedules/systemd/index.html#how-to-change-the-default-systemd-unit-and-timer-file-using-a-template

Hm. I've thought about doing this, but I don't see how I can use different values for different profiles, because the template is a global setting. So, if I have one profile running every 10 minutes, I might want to add a 2 minute randomized delay to it, but if I have another profile running every 24 hours, 1 _hour_ of randomized delay might make more sense there. Is there a way to do this using the custom template?

I even might want to use different delays for different _tasks_ in the same profile. For example, run `backup` every 10 minutes (± 2 minutes), but run `forget` only every 24 hours (± 2 hours).

[...]

PR #340, opened 2024-03-15
drop-ins: transparent timer drop-in support

jkellerer:
PR for #217 (and other systemd specific configs for timers), tested in unit and manual tests.

Note: This also contains a fix for run-schedule not transferring `AfterNetworkOnline` & `SystemdDropInFiles` and a weakness of the drop-in support not detecting the own files when the sources use an extension other than `.conf`.


creativeprojects (approved):
LGTM

Thanks for taking care of it so quickly 😉

PR #342, opened 2024-03-16
docs: add note about missing syslog on Windows

omoe61:
I couldn't find any hint in the docs that the syslog destination isn't available on Windows. Therefore I added a note to "Configuration -> Logs".


creativeprojects:
It's a good  point, it was definitely missing from the documentation, thanks for the PR 👍🏻

PR #343, opened 2024-03-16
log: allow controlling command output redirection

jkellerer:
This PR is a super-light version of outstanding PR #207 to control when command output is sent to logs, console or both, with a default of "`auto`" so that interactive usage isn't fully silent when log is specified (e.g. in global settings).

```
root@log:/# ./resticprofile --log=test.log containers.check
using temporary cache in /tmp/restic-check-cache-3409750633
create exclusive lock for repository
load indexes
check all packs
check snapshots, trees and blobs
[3:13] 100.00%  580 / 580 snapshots

no errors were found
```

```
root@log:/# cat test.log 
2024/03/16 22:12:12 INFO  using configuration file: /etc/resticprofile/profiles.yaml
2024/03/16 22:12:12 INFO  profile 'containers': initializing repository (if not existing)
2024/03/16 22:12:12 INFO  profile 'containers': starting 'check'
using temporary cache in /tmp/restic-check-cache-3409750633
create exclusive lock for repository
load indexes
check all packs
check snapshots, trees and blobs
[3:13] 100.00%  580 / 580 snapshots

no errors were found
```

.. PR #207 is still required to have proper terminal output in this case but it is better than none.


jkellerer:
@creativeprojects, let me know what you think about this and whether names need to be adjusted.

creativeprojects:
Yes, this is a very good idea.

I'm not too sure about the name of the flag either. We could also do:
- `log-output` (with something like `dual`, `log-only`) ?
- `log-console` (boolean) ?

🤔

jkellerer:
name was also one thing were I wasn't sure yet. Had used `log-commands` from `default-command`.

Maybe `add-output-to` (e.g. `--add-output-to=log`, `--add-output-to=console`, `--add-output-to=both`)

Mostly the flag wont be needed as the `auto` value should be fine. Also we have some commands where the output must not be sent to logs. It may be good to have a name that doesn't get in the way if output cannot be added.

jkellerer:
Changed the name to "command-output", kind of a combination of yours and mine proposals :)

```
resticprofile --log=rp.log --command-output=log ...
resticprofile --log=rp.log --command-output=console ...
resticprofile --log=rp.log --command-output=all ...
resticprofile --log=rp.log --command-output=auto ...

# also valid
resticprofile --log=rp.log --command-output=log,console ... 
```

PR #344, opened 2024-03-16
syslog: local syslog and stdout redirection

jkellerer:
In relation to #343, this PR adds local syslog server and command output redirection into syslog.

Motivation: Currently I redirect systemd output to syslog (including all stdout/err from resticprofile). This PR allows to have the same support using the built-in syslog client, which has the advantage of more consistent timestamps (not duplicated in stdout formatted log).

The following additional log URLs are added:
* `syslog:` (local syslog server via unix socket)
* `syslog://hostname` (udp on default port 514)
* `syslog://hostname:514` (udp on port 514)
* `syslog-tcp://hostname` (tcp on default port 514)
* `syslog-tcp://hostname:514` (tcp on port 514)


jkellerer:
@creativeprojects , do you have any concerns with this. It does change previous behaviour in that it now sends command output by default (though #343 allows to control it).

On the other hand the previous version was a bit inconsistent since we sent it to log files but not to syslog.

creativeprojects:
Yeah I'm fine with it 👍🏻 
Thanks for the PR 😉

creativeprojects (approved):
Works fine, and makes more sense than the previous implementation 👍🏻 

Thanks 😉

PR #346, opened 2024-03-19
contrib: added the common mixins library

jkellerer:
This PR is a contribution of a common mixins library that may be used to simplify common actions when performing backups. It starts with support for snapshots and databases in posix environments.

Status:

- [ ] integration tested
- [ ] tested

Status for the moment, it is just a brain dump of something I wanted to add for a couple of time but it is highly untested, though I use most of the actions in my backups in simplified forms.

### DIFF
--- a/contrib/mixins/README.md
+++ b/contrib/mixins/README.md
@@ -1,0 +1,65 @@
+# Common mixins library
+
+A library of reusable mixins to simplify common tasks needed when taking backups.
+
+Currently, this library contains mixins for:
+* Dumping databases to restic stdin (mysql, mariadb, pgsql)
+* Taking temporary snapshots (lvm, btrfs)
+* Temporarily freezing VM disk images (virsh/kvm)
+
+## Example Usage
+
+`profiles.toml`
+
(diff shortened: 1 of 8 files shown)


creativeprojects:
This is an excellent idea 💡 

I always wanted to add a section on how to use `rclone` with `restic` with a few configuration examples, but never took the time to do it 😞

creativeprojects:
BTW if we want the various snippets to also work with Windows, we can do like [go-task](https://taskfile.dev/#features) and use [a shell interpreter](https://github.com/mvdan/sh)

This is out of the scope of this MR though

jkellerer:
Was actually thinking to extend the shell hook format with an optional struct to toggle things like: 

* what shell to use
* async (run until the section ends, e.g. for remote locks with external tools)
* hidden-execution (copy to script file to hide confidentials from proc list) 
* what to do on return code (abort, fail, continue).. allows to use it as conditional when set to abort instead of fail.

...also out of scope for this PR. But helpful here and an integration point for a shell interpreter.

PR #350, opened 2024-03-23
add "stdin-from-command" (0.17) and minors of restic 0.16

jkellerer:
Maybe a bit early, prepares resticprofile for restic 0.17


jloh:
Anything I can do to help get this across the line? Would love to be able to use `skip-if-unchanged` in my backups 👍

creativeprojects:
I think @jkellerer is busy with other things right now. I haven't had a look at the PR in details but I know there are a few things to update to support the new `--stdin-from-command` backup option. I'll have a look at it soon-ish 😉

creativeprojects:
We're getting some shell escaping issue while trying to move our implementation to the new `--stdin-from-command` flag. I'm not going to leave our `stdin-command` implementation for now and just build with support for all the new 0.17 flags for now.

PR #353, opened 2024-03-28
Add "--stderr" to redirect console to stderr (for "cat" and "dump")

### LINKED ISSUE #352: resticprofile writes log information to stdout when using restic dump
I have been dumping postgresql databases into restic using stdin.  

When I attempt to restore these backups, I am using the restic dump command to send the dump file contents to standard output and piping it to pg_restore as follows:

```bash
resticprofile profile.dump --path /sql.pgdump latest /sql.pgdump | pg_restore -d dbname
```

But the beginning of the stanard output stream includes the following 2 lines generated by resticprofile ahead of the actual dump file contents:

```
2024/03/28 20:11:37 using configuration file: /etc/resticprofile/profiles.toml  
2024/03/28 20:11:37 profile 'db': starting 'dump'                               
PGDM |dbname#14.6 (Ubuntu 14.6-0ubuntu0.22.04.1)#14.6 (Ubuntu 14.6-0ubuntu0.22.0
0ENCODINENCODINGSET client_encoding = 'UTF8';
...
```

pg_restore is complaining about the extraneous output generated by resticprofile.  I [initially reported this issue,](https://github.com/creativeprojects/resticprofile/issues/153#issuecomment-1676495630) but probably not as clearly as shown above.

I propose, when using the restic dump command with resticprofile, either logging is suppressed on standard output entirely, or is instead written to standard error.  

The new run hooks don't really help here when you want to send the dump output to a variety of different commands.  Or in my case, I don't want to accidentally pipe the output directly to a db restore and wipe anything out.  

What do you think?

jkellerer:
This PR fixes #352 and allows to redirect console output to stderr. Also it auto-enables it for "dump" and "cat" commands.

E.g.:

```shell
./resticprofile dump latest /opt/test/Dockerfile > restored-Dockerfile

2024/03/28 16:32:24 using configuration file: /etc/resticprofile/profiles.yaml
2024/03/28 16:32:24 profile 'default': initializing repository (if not existing)
2024/03/28 16:32:25 profile 'default': starting 'dump'
2024/03/28 16:32:34 profile 'default': finished 'dump'
```

`cat restored-Dockerfile`:

```Dockerfile
FROM buildpack-deps/bullseye 
 
ENV VERSION=7ff74b2d5288fbc35c4f6fcb94466d408b1f853c 
ENV UID=1000 
 
RUN apt-get update \ 
 && apt-get install -y \
...
```

Note: Currently this doesn't cover output of own commands, however I think we can keep it simple for the moment and ignore this limitation.

### DIFF
--- a/constants/command.go
+++ b/constants/command.go
@@ -10,6 +10,7 @@ const (
 	CommandSnapshots = "snapshots"
 	CommandUnlock    = "unlock"
 	CommandMount     = "mount"
+	CommandCat       = "cat"
 	CommandCopy      = "copy"
 	CommandDump      = "dump"
 	CommandFind      = "find"
--- a/docs/content/usage/_index.md
+++ b/docs/content/usage/_index.md
@@ -82,6 +82,7 @@ There are not many options on the command line, most of the options are in the c
 * **[-v | --verbose]**: Force resticprofile and restic to be verbose (override any configuration from the profile)
 * **[--trace]**: Display even more debugging information
 * **[--no-ansi]**: Disable console colouring (to save output into a log file)
+* **[--stderr]**: Send console output from resticprofile to stderr (is enabled for commands `cat` and `dump`)
 * **[--no-lock]**: Disable resticprofile locks, neither create nor fail on a lock. restic locks are unaffected by this option.
 * **[--theme]**: Can be `light`, `dark` or `none`. The colours will adjust to a 
 light or dark terminal (none to disable colouring)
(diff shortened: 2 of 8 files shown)


creativeprojects (commented):
There is something funny going on:

If you run a `dump` or `cat` command without redirection, you get a coloured `stderr` and the content displayed.

If you redirect the output to a pipe or file, then we detect that `stdout` is not a terminal and `stderr` is displayed without colours.

It's certainly not a big deal, but I'm wondering if we wouldn't get some other corner cases because of it?

I also can't remember where the terminal detection is done ❓ 🤔

jkellerer:
Had also seen the colouring issue. That is caused in the `color` package because it doesn't handle stderr and stdout differently. If stdout is redirected, colors are disabled. We could potentially fix it by overriding the `color.NoColor` flag but this has to be done very carefully therefore I think we should leave it as it is.

creativeprojects (approved):
Right that's the color package doing it. I was wondering about that since I couldn't find it anywhere in the code.

Looks good to me 😄 

Thanks for taking care of it 👍🏻

PR #354, opened 2024-03-29
Add option to set working directory for restic backup

seiuneko:
This PR introduces the `change-working-dir` option within the resticprofile configuration, enabling users to set a specific working directory for the restic backup command. This enhancement is particularly beneficial for users backing up Btrfs snapshots, where the ability to use relative paths is crucial. By utilizing relative paths, we can omit the snapshot path prefix in the restic repository, which significantly simplifies the restoration process from snapshots.

Prior to this change, resticprofile would always expand relative paths to absolute paths (e.g., converting `source: .` to its absolute path), and the `base-dir` setting was only modifiable at the profile level, affecting path resolution across different sections.

With the addition of the `change-working-dir` option, users now have the flexibility to ensure that the `source` paths are not expanded into absolute paths. This is achieved by changing the restic backup command's working directory to the one specified by `source-base`.

I'm relatively new to Golang, so if there's any aspect of my implementation that's incorrect or if I've overlooked anything, please do let me know. I appreciate your feedback and thank you in advance!

### DIFF
--- a/config/profile.go
+++ b/config/profile.go
@@ -654,7 +654,7 @@ func (p *Profile) resolveSourcePath(sourceBase string, sourcePaths ...string) []
 
 	sourceBase = fixPath(sourceBase, expandEnv, expandUserHome)
 	// When "source-relative" is set, the source paths are relative to the "source-base"
-	if !p.Backup.SourceRelative {
+	if p.Backup == nil || !p.Backup.SourceRelative {
 		// Backup source is NOT relative to the configuration, but to PWD or sourceBase (if not empty)
 		// Applying "sourceBase" if set
 		if sourceBase = strings.TrimSpace(sourceBase); sourceBase != "" {
--- a/config/profile.go
+++ b/config/profile.go
@@ -172,6 +172,7 @@ type BackupSection struct {
 	CheckAfter                       bool     `mapstructure:"check-after" description:"Check the repository after the backup command succeeded"`
 	UseStdin                         bool     `mapstructure:"stdin" argument:"stdin"`
 	StdinCommand                     []string `mapstructure:"stdin-command" description:"Shell command(s) that generate content to redirect into the stdin of restic. When set, the flag \"stdin\" is always set to \"true\"."`
+	ChangeWorkingDir                 bool     `mapstructure:"change-working-dir" description:"Change the restic backup command's working directory to \"source-base\". When set, the \"source\" does not expand to absolute paths"`
 	SourceBase                       string   `mapstructure:"source-base" examples:"/;$PWD;C:\\;%cd%" description:"The base path to resolve relative backup paths against. Defaults to current directory if unset or empty (see also \"base-dir\" in profile)"`
 	Source                           []string `mapstructure:"source" examples:"/opt/;/home/user/;C:\\Users\\User\\Documents" description:"The paths to backup"`
 	Exclude                          []string `mapstructure:"exclude" argument:"exclude" argument-type:"no-glob"`
@@ -651,15 +652,19 @@ func (p *Profile) SetRootPath(rootPath string) {
 func (p *Profile) resolveSourcePath(sourceBase string, sourcePaths ...string) []string {
 	var applySourceBase, applyBaseDir pathFix
 
-	// Backup source is NOT relative to the configuration, but to PWD or sourceBase (if not empty)
-	// Applying "sourceBase" if set
-	if sourceBase = strings.TrimSpace(sourceBase); sourceBase != "" {
-		sourceBase = fixPath(sourceBase, expandEnv, expandUserHome)
-		applySourceBase = absolutePrefix(sourceBase)
-	}
-	// Applying a custom PWD eagerly so that own commands (e.g. "show") display correct paths
-	if p.BaseDir != "" {
-		applyBaseDir = absolutePrefix(p.BaseDir)
+	sourceBase = fixPath(sourceBase, expandEnv, expandUserHome)
+	// When "change-working-dir" is set, the source paths are relative to the "source-base"
+	if !p.Backup.ChangeWorkingDir {
+		// Backup source is NOT relative to the configuration, but to PWD or sourceBase (if not empty)
+		// Applying "sourceBase" if set
+		if sourceBase = strings.TrimSpace(sourceBase); sourceBase != "" {
+			applySourceBase = absolutePrefix(sourceBase)
+		}
+		// Applying a custom PWD eagerly so that own commands (e.g. "show") display correct paths
+		if p.BaseDir != "" {
+			applyBaseDir = absolutePrefix(p.BaseDir)
+		}
+
 	}
 
 	// prefix paths starting with "-" with a "./" to distinguish a source path from a flag
--- a/config/profile.go
+++ b/config/profile.go
@@ -199,7 +199,7 @@ func (b *BackupSection) resolve(profile *Profile) {
 	if b.unresolvedSource == nil {
 		b.unresolvedSource = b.Source
 	}
-	b.Source = profile.resolveSourcePath(b.SourceBase, b.unresolvedSource...)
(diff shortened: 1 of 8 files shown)


jkellerer:
@seiuneko , thanks a lot for your contribution 👍 at first sight it looks correct to me but I need to test it. So far there was the assumption that restic doesn't store relative paths but looking at it again it looks that in fact it does store the information internally (though not visible when listing snapshots) and it does make a difference for restore. Thanks for pointing it out.

Speaking of the config option, since what we really want is to enable relative source dir, I imagine the property could be named `SourceRelative` (the fact that it changes CWD for restic is a technical requirement so it may be documented but the main point is that source paths are kept relative).

shell/command_test.go
```diff
@@ -285,6 +285,23 @@ func TestSelectCustomShell(t *testing.T) {
 	assert.Empty(t, shell)
 }
 
+func TestRunShellWorkingDir(t *testing.T) {
+	buffer := &bytes.Buffer{}
```
jkellerer, line 289: `buffer := new(strings.Builder)` can be used for `cmd.Stdout` and has `buffer.String()`

shell/command_test.go
```diff
@@ -285,6 +285,23 @@ func TestSelectCustomShell(t *testing.T) {
...
+	cmd := NewCommand("pwd", nil)
+	cmd.Stdout = buffer
+	cmd.Dir = "/tmp"
```
jkellerer, line 292: we might want to use `t.TempDir` to not depend on the platform

shell/command_test.go
```diff
@@ -285,6 +285,23 @@ func TestSelectCustomShell(t *testing.T) {
 	assert.Empty(t, shell)
 }
 
+func TestRunShellWorkingDir(t *testing.T) {
+	buffer := &bytes.Buffer{}
+	cmd := NewCommand("pwd", nil)
```
jkellerer, line 290: For `platform.IsWindows()` the command is `@echo %CD%` (for cmd.exe) or `(pwd).Path` for powershell. Maybe we can get away with skipping the test for the platform.

config/profile.go
```diff
@@ -651,15 +652,19 @@ func (p *Profile) SetRootPath(rootPath string) {
 func (p *Profile) resolveSourcePath(sourceBase string, sourcePaths ...string) []string {
 	var applySourceBase, applyBaseDir pathFix
 
-	// Backup source is NOT relative to the configuration, but to PWD or sourceBase (if not empty)
-	// Applying "sourceBase" if set
-	if sourceBase = strings.TrimSpace(sourceBase); sourceBase != "" {
```
jkellerer, line 656: I see an issue with this change. Originally all actions happened when trimmed sourceBase base not empty. With the new logic, sourceBase is formatted and may have a value when it was set to a space char.

To avoid side effects, I'd prefer to leave the original logic and instead set `applySourceBase` and `applyBase` to `nil` when the new option is set.

config/profile.go
```diff
@@ -651,15 +652,19 @@ func (p *Profile) SetRootPath(rootPath string) {
...
-		sourceBase = fixPath(sourceBase, expandEnv, expandUserHome)
-		applySourceBase = absolutePrefix(sourceBase)
-	}
-	// Applying a custom PWD eagerly so that own commands (e.g. "show") display correct paths
-	if p.BaseDir != "" {
-		applyBaseDir = absolutePrefix(p.BaseDir)
+	sourceBase = fixPath(sourceBase, expandEnv, expandUserHome)
+	// When "change-working-dir" is set, the source paths are relative to the "source-base"
+	if !p.Backup.ChangeWorkingDir {
```
jkellerer, line 657: `p.Backup` should be checked for `nil`, e.g. if applying all other comments:

```go
if p.Backup != nil && p.Backup.SourceRelative {
    applySourceBase = nil
    applyBaseDir = nil
}
```

At the moment it wont be nil when this is used but a nil check in this case would be better as the callers of the method can legally use it when the backup section is missing.

config/profile.go
jkellerer: Ok.. need to admit looking at the entire method this side effect does not matter as only the `apply*` are used. In that case just wrap the entire block into the if condition (without modification).

shell/command_test.go
seiuneko: According to the logs from GitHub Actions on the Windows platform, the command `make test-ci` was executed using PowerShell, so I opted for the `(pwd).Path` command.

seiuneko:
@jkellerer Thank you for your thorough review. I have revised the code according to your suggestions. Are there any other issues?

config/profile.go
```diff
@@ -651,15 +652,19 @@ func (p *Profile) SetRootPath(rootPath string) {
...
+		// Applying "sourceBase" if set
+		if sourceBase = strings.TrimSpace(sourceBase); sourceBase != "" {
+			applySourceBase = absolutePrefix(sourceBase)
+		}
+		// Applying a custom PWD eagerly so that own commands (e.g. "show") display correct paths
+		if p.BaseDir != "" {
+			applyBaseDir = absolutePrefix(p.BaseDir)
+		}
+
 	}
```
jkellerer, line 668: At this step we should add the following: 

```go
} else if p.BaseDir == "" && sourceBase == "" && p.config != nil {
    p.config.reportChangedPath(".", "", "source-base (for relative source)")
}
```

This prints a warning when nether `source-base` nor `base-dir` is defined but paths are set to relative.

config/profile.go
```diff
@@ -651,15 +652,19 @@ func (p *Profile) SetRootPath(rootPath string) {
...
-	if sourceBase = strings.TrimSpace(sourceBase); sourceBase != "" {
-		sourceBase = fixPath(sourceBase, expandEnv, expandUserHome)
-		applySourceBase = absolutePrefix(sourceBase)
-	}
-	// Applying a custom PWD eagerly so that own commands (e.g. "show") display correct paths
-	if p.BaseDir != "" {
-		applyBaseDir = absolutePrefix(p.BaseDir)
+	sourceBase = fixPath(sourceBase, expandEnv, expandUserHome)
+	// When "source-relative" is set, the source paths are relative to the "source-base"
+	if p.Backup == nil || !p.Backup.SourceRelative {
```
jkellerer, line 657: After checking it further, I can see we need an additional condition here. The method is used to resolve backup sources in the backup section and also in `SetPath`. When called in `SetPath`, it must not make the paths relative because `SetPath` sets the path args for non-backup commands and they expect absolute paths.

I suggest we change the method signature, from:

```go
func (p *Profile) resolveSourcePath(sourceBase string, sourcePaths ...string) []string {
```

to: 

```go
func (p *Profile) resolveSourcePath(sourceBase string, relativePaths bool, sourcePaths ...string) []string {
```

This also simplifies the if condition since the only caller where this is allowed to be `true` is `p.Backup`.

jkellerer:
LGTM

Thanks a lot for the updates, just tried it with the following profile and it works perfectly fine (including error reporting for missing base paths):

```yaml
relative:
  inherit: default
  base-dir: /opt/containers

  backup:
    source-relative: true
    source-base: /opt/snapshot_containers
    source: 
      - test-container
    run-before:
      - 'btrfs subvolume snapshot -r /opt/containers /opt/snapshot_containers'
    run-finally:
      - 'btrfs subvolume delete /opt/snapshot_containers'

  ls:
    path: true

  restore:
    path: true
    target: '.'
```

In detail:

* `resticprofile relative.backup` - create a backup from btrfs snapshopt
* `resticprofile relative.ls latest` - lists the files (with relative paths)
* `resticprofile relative.restore latest` - restores files into the original base-dir

Only remaining issue I found is that restic resolves paths differently than resticprofile but I'll take care of it in a separate PR.

jkellerer:
@creativeprojects , I'm good with the contribution, do you have any concerns?

creativeprojects:
Thanks for the PR, I'll give it a try on my zfs snapshots 👍🏻

creativeprojects:
I guess it doesn't really change much in the end, but it can simplify the configuration.

It looks good to me 👍🏻

PR #366, opened 2024-06-04
Add documentation on how to ingest resticprofile stats into telegraf

yacoob:
I've spent some time experimenting with JSON parsing, thought I'd add this here :)


contrib/telegraf/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+1. Configure [http_listener_v2](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/http_listener_v2) expecting Prometheus data, have resticprofile push to it via `prometheus-push`. This is currently slightly borked, due to [this bug](https://github.com/influxdata/telegraf/issues/15453). The data gets written, but resticprofile emits a warning. Once that bug is fixed, it might be a viable way to go. The downside is that you need to explicitly configure urls on telegraf side, like so:
+
+```toml
+[[inputs.http_listener_v2]]
+  paths = ["/metrics/job/home.backup"]
+  service_address = "127.0.0.1:9999"
+  data_format = "prometheus"
+```
```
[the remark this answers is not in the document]
yacoob: There's no workaround needed here, just use the first approach :)

contrib/telegraf/README.md
```diff
@@ -0,0 +1,30 @@
+# backup stats in InfluxDB via Telegraf
+
+Telegraf can ingest stats produced by resticprofile in three different ways:
+
+1. Configure Telegraf to parse resticprofile's JSON status file. The parsing directive looks brittle, but as of June 2024, this seems to be the most viable method.
```
creativeprojects, line 5: Could you please elaborate why the parsing directive looks brittle?

(for information we're not planning on changing the format of the status file anytime soon)

creativeprojects (approved):
Thank you very much for sharing this information 👍🏻

contrib/telegraf/README.md
yacoob: It's more of a gut feeling on my side, powered by the amount of time spent on that `(...).@group` solution :)

More seriously though, it does encode expectations how that JSON structure will look like on the consuming side. It's one more place to worry about if that format ever changes.

yacoob:
I've updated the instructions, as telegraf 1.31 is out as of today.

PR #393, opened 2024-08-11
docs: add info for installing using winget

danielfdickinson:
winget comes with current version of Windows 10 and 11 and provides command-line installation of software for Windows.

### DIFF
--- a/docs/content/installation/windows.md
+++ b/docs/content/installation/windows.md
@@ -24,6 +24,14 @@ Resticprofile can be installed from [scoop](https://scoop.sh) main bucket:
 scoop install resticprofile
 ```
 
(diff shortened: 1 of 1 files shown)


creativeprojects:
Nice! I didn't know it was available on winget 😃 

Thanks for the information 👍🏻

creativeprojects:
### Note to self
It looks like it can be automated: a release pipeline could open a PR in https://github.com/microsoft/winget-pkgs

danielfdickinson:
> ### Note to self
> 
> It looks like it can be automated: a release pipeline could open a PR in https://github.com/microsoft/winget-pkgs

That would be awesome! And, glad the docs could help.

PR #405, opened 2024-09-29
feat(bash-completion): prevent space after profile completion `.`

myii:
When tab completing a profile, an unwanted space is added after the period `.`, necessitating hitting backspace before further completions can be accessed.

The implementation here was based on the following resources:

* https://stackoverflow.com/a/66151065
  - Add spaces to the end of some Bash autocomplete options, but not to others?
* https://stackoverflow.com/a/19271994
  - How can I check the last character in a string in bash?

An alternative to checking for the last character is using regex:

* https://stackoverflow.com/a/21425089
  - How to check if the last string character equals '*' in Bash?


creativeprojects:
Hey!

Sorry it took me a **very** long time to finally look into how shell completion works (it was all done by someone else). Also I'm not using `zsh` completion myself 😬 

I gave it a try with `bash` and it does work nicely 👍🏻 

I didn't understand why it was breaking the `zsh` completion, hence why I left your PR on the side. I think I understand how it works now so I'll rewrite the `zsh` completion in another PR.

Thank you for this PR 👍🏻

PR #478, opened 2025-03-30
[Docs] Correct configuration search paths on Windows

auxym:
Append `resticprofile` to the AppData and ProgramData paths on Windows.

See https://github.com/creativeprojects/resticprofile/blob/ac3b810f2f735642d0a0cea8ec46c585b80904bc/filesearch/filesearch.go#L224, resticprofile adds its AppName to the config dir paths.

Also add `%USERPROFILE%\AppData\Roaming\resticprofile`, which should be included in the paths returned by `xdg.ConfigDirs` (see https://github.com/adrg/xdg/blob/d8917aeea69ad0645ce98b2a1e3edb5cab67199a/paths_windows.go#L22).

Finally, I added a note to clarify that these paths can be overridden if the XDG environment variables are set.

### DIFF
--- a/docs/content/configuration/path.md
+++ b/docs/content/configuration/path.md
@@ -153,8 +153,12 @@ resticprofile will search for your configuration file in these folders:
 
 resticprofile will search for your configuration file in these folders:
 - _current directory_
-- %USERPROFILE%\AppData\Local\
-- c:\ProgramData\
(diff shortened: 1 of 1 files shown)


creativeprojects:
Well spotted, the documentation was definitely outdated.

Thank you very much for the PR 👍🏻 

(the documentation will be updated on the next release, possibly today or in the next few days)

PR #486, opened 2025-04-14
Add copy from repository in configuration

hgraeber:
Fix the copy command so that `from-repository` and `from-repository-file` works like they work with restic alone. That means
```yaml
version: 2
profile:
  repository: r1
  ...
  copy:
    from-repository: r2
    ...
```
copies from repository `r2` to `r1` and not in the opposite direction, like it does now. That means the repository mentioned directly below `profile` (`r1`) is the destination similar to `backup`. The special case without the "from"-prefix mention in the docs works like before.

As a bonus the retention section can allow  `before-copy` and `after-copy` analog to `before-backup` and `after-backup`. This will be reserved for a future PR.

### DIFF
--- a/config/profile.go
+++ b/config/profile.go
@@ -344,96 +344,156 @@ func (s *ScheduleBaseSection) getScheduleConfig(p *Profile, command string) *Sch
 	return s.scheduleConfig
 }
 
-// CopySection contains the destination parameters for a copy command
+// CopySection contains the source or destination parameters for a copy command
 type CopySection struct {
 	GenericSectionWithSchedule  `mapstructure:",squash"`
 	Initialize                  bool              `mapstructure:"initialize" description:"Initialize the secondary repository if missing"`
 	InitializeCopyChunkerParams maybe.Bool        `mapstructure:"initialize-copy-chunker-params" default:"true" description:"Copy chunker parameters when initializing the secondary repository"`
-	Repository                  ConfidentialValue `mapstructure:"repository" description:"Destination repository to copy snapshots to"`
-	RepositoryFile              string            `mapstructure:"repository-file" description:"File from which to read the destination repository location to copy snapshots to"`
-	PasswordFile                string            `mapstructure:"password-file" description:"File to read the destination repository password from"`
-	PasswordCommand             string            `mapstructure:"password-command" description:"Shell command to obtain the destination repository password from"`
-	KeyHint                     string            `mapstructure:"key-hint" description:"Key ID of key to try decrypting the destination repository first"`
+	FromRepository              ConfidentialValue `mapstructure:"from-repo" argument:"from-repo" description:"Source repository to copy snapshots from"`
+	FromRepositoryFile          string            `mapstructure:"from-repository-file" argument:"from-repository-file" description:"File from which to read the source repository location to copy snapshots from"`
+	FromPasswordFile            string            `mapstructure:"from-password-file" argument:"from-password-file" description:"File to read the source repository password from"`
+	FromPasswordCommand         string            `mapstructure:"from-password-command" argument:"from-password-command" description:"Shell command to obtain the source repository password from"`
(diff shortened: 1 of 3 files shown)


creativeprojects:
Hey! how is it going with this PR?

Tell me if you need any help 👍🏻

hgraeber:
It's the part of `before-copy` and `after-copy` that's missing (The same for `check-before` and `check-after`). I have made already two tries I never pushed, because I don't like the resulting code.

`copy` with `from-repository` is already working. Maybe I shall make it ready for release and leave the `before-copy`/`after-copy` part for another PR.

creativeprojects:
If the first part of the PR is ready, you can definitely split it in two.
Smaller PR always makes the code quicker to review 👍🏻

config/profile.go
```diff
@@ -344,96 +344,156 @@ func (s *ScheduleBaseSection) getScheduleConfig(p *Profile, command string) *Sch
...
 	InitializeCopyChunkerParams maybe.Bool        `mapstructure:"initialize-copy-chunker-params" default:"true" description:"Copy chunker parameters when initializing the secondary repository"`
-	Repository                  ConfidentialValue `mapstructure:"repository" description:"Destination repository to copy snapshots to"`
-	RepositoryFile              string            `mapstructure:"repository-file" description:"File from which to read the destination repository location to copy snapshots to"`
-	PasswordFile                string            `mapstructure:"password-file" description:"File to read the destination repository password from"`
-	PasswordCommand             string            `mapstructure:"password-command" description:"Shell command to obtain the destination repository password from"`
-	KeyHint                     string            `mapstructure:"key-hint" description:"Key ID of key to try decrypting the destination repository first"`
+	FromRepository              ConfidentialValue `mapstructure:"from-repo" argument:"from-repo" description:"Source repository to copy snapshots from"`
```
creativeprojects: Shouldn't we call it `from-repository` for consistency with the other configuration options?

(I know restic calls it `from-repo` and that was a poor design decision at the start of the resticprofile project)

hgraeber: You are right it shall be `from-repository`. That's consistent with the `init` section, too.

hgraeber: Changed that...

creativeprojects:
Thanks for the change 👍🏻 

Actually I found something during testing: when the `initialise: true` flag is set, we're trying to initialise the target repository if it's not existing yet. Which makes sense in a way that it's going to be a copy of the current one.

While leaving the `initialise: true` flag using a `from-repository` it doesn't try to initialise the source repository. And it kind of make sense as well since we should be copying from an existing repository.

```
2025/07/30 18:17:59 profile 'self': initializing repository (if not existing)
2025/07/30 18:17:59 starting command: /usr/local/bin/restic init --password-file=examples/key --repo=/Volumes/RAMDisk/self --verbose=1
2025/07/30 18:17:59 profile 'self': initializing secondary repository (if not existing)
2025/07/30 18:17:59 starting command: /usr/local/bin/restic init --copy-chunker-params --from-password-file=examples/key --from-repo=/Volumes/RAMDisk/copy-from --password-file=examples/key --repo=/Volumes/RAMDisk/self --verbose=1
```

So we have two choices:
- we can disable trying to initialise the secondary (source) repository
- we can be sending the source repository to the secondary init to fix the initialisation step: it makes things consistent (but not very useful)

Any preference?

creativeprojects:
... which I realise is also opening yet another a can of worms 😆 

In the order of initialisation, the source should be done first, then the target with the `--copy-chunker-params` flag so we'd need to change the order: right now the `repository` is done first, like any other step, then the one from the `copy` section 🤦🏻

hgraeber:
It is important to initialize the target repository to be consistent with the backup section. This is useful for making copies to external drives and initialize them on first use. Nor forgetting to use `--copy-chunker-params` is important here.

I wouldn't bother with initializing a source drive. This is really an edge case. Possibly forgetting `--copy-chunker-params` doesn't count here, because there is no source repo for this, which makes sense.

creativeprojects:
Makes sense 😉 
Right, I'll merge this first PR for now, thanks for the great work! 👍🏻

PR #488, opened 2025-04-15
Add functionality to upload log-file

FH3095:
Hello,
I use restic to backup different machines, including some windows machines.
To monitor for any problems, I like to receive the log-files on some central server to check for problems.
Since on windows syslog is not supported, that sadly is not an option.
So I added a config option to upload the log-file to some configured URL.
Because I think this could be useful for somebody else I opened this pull request.

Greetings
FH


creativeprojects:
Hey!

Thanks for this PR, I think it will be useful indeed.

The issue in the tests are fixed: you need to rebase your branch from `master` and that should work 👍🏻

Issue #485, opened 2025-04-10
used restic version in metrics and/or logs


rainerrose:
Hi!

For debugging purposes, it would be very helpful if I could see the restic version used in the backup run.

It would also be nice if this information could optionally be included in the log as well as in the prometheus statistics.
Or I could add this information via `prometheus-labels`.

The only possibility I can think of at the moment would be a script as a command hook (e.g. run-before). But this is not really nicely solved in my opinion and not very compatible with the Prometheus syntax output. (E.g. if I would use `restic --json version`).
In addition, I have no access to the possibly different location that may be defined under:
~~~yaml
global:
  restic-binary: /foo/bar/restic
~~~

I suspect that all this is easier to solve in the resticprofile environment, so I have created this issue as a feature request.

creativeprojects:
You can get this information using the `-v` or `--verbose` flag (and admittedly you'll also get a lot of other messages as well)

```
% resticprofile -n azure-dev -v
2025/04/10 20:35:29 resticprofile 0.30.0 compiled with go1.24.2
2025/04/10 20:35:29 using configuration file: profiles.yaml
2025/04/10 20:35:29 loading: profiles.yaml
2025/04/10 20:35:29 memory available: 14939MB
2025/04/10 20:35:29 using restic 0.18.0
2025/04/10 20:35:29 profile 'azure-dev': starting 'snapshots'
[...]
```

I suppose I can add it to the prometheus labels 😉

rainerrose:
Hm.
From this I conclude that the information is already available.
Wouldn't it be an idea to have this info available as a metric similar to `resticprofile_build_info`?
With `prometheus-labels` this info IMHO would appear in every metric line.

creativeprojects:
I'm not actively using Prometheus, by that I mean I don't normally write queries, build graphs or things like that.

So, does it make sense to have a new metric like that?

```
[...]

# TYPE restic_build_info gauge
resticprofile_build_info{goversion="go1.24.1",profile="prom",version="0.18.0"} 1

# TYPE resticprofile_build_info gauge
resticprofile_build_info{goversion="go1.24.2",profile="prom",version="0.30.0"} 1
```

rainerrose:
> So, does it make sense to have a new metric like that?

Oh yes, that looks cool; that would be great.

creativeprojects:
Merged in #495 
Will be included in next release `v0.31.0`

PR #495, opened 2025-04-23
add version of restic to the Prometheus metrics

antondollmaier:
Add the restic version as a new label to the existing metric `resticprofile_build_info`.

This allows tracking the versions of both restic and resticprofile: PromQL can be used to identify outdated installations.

I'm not an experienced developer in Go, so there might be better ways to solve this.

Adding the new metric `restic_build_info` failed for unknown reasons (the metric didn't show up in the output file).

refs #485

### DIFF
--- a/monitor/prom/metrics.go
+++ b/monitor/prom/metrics.go
@@ -22,13 +22,14 @@ const (
 )
 
 type Metrics struct {
-	labels   prometheus.Labels
-	registry *prometheus.Registry
-	info     *prometheus.GaugeVec
-	backup   BackupMetrics
+	labels     prometheus.Labels
+	registry   *prometheus.Registry
+	info       *prometheus.GaugeVec
+	resticInfo *prometheus.GaugeVec
+	backup     BackupMetrics
 }
 
-func NewMetrics(profile, group, version string, configLabels map[string]string) *Metrics {
+func NewMetrics(profile, group, version string, resticversion string, configLabels map[string]string) *Metrics {
 	// default labels for all metrics
 	labels := prometheus.Labels{profileLabel: profile}
 	if group != "" {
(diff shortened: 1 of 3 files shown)


creativeprojects:
@antondollmaier thanks for this PR 👍🏻 

Does it make more sense to create a new metric called `restic_build_info` than adding a label to the existing one?

I'm asking because it shouldn't be difficult to add a new metric, if that makes things easier to query from prometheus

antondollmaier:
Thanks for the input!

> Does it make more sense to create a new metric called `restic_build_info` than adding a label to the existing one?
> 
> I'm asking because it shouldn't be difficult to add a new metric, if that makes things easier to query from prometheus

I initially failed to do so, but succeeded now. And yes, the `restic_build_info` does make more sense than extending the resticprofile metric with another label.

The metrics now look like this:

~~~text
# HELP restic_build_info restic build information.
# TYPE restic_build_info gauge
restic_build_info{profile="laptop",version="0.16.4"} 1
# HELP resticprofile_build_info resticprofile build information.
# TYPE resticprofile_build_info gauge
resticprofile_build_info{goversion="go1.24.2",profile="laptop",version="0.30.0-dev"} 1
~~~

Is that okay for you as well?

monitor/prom/metrics.go
```diff
@@ -13,22 +13,24 @@ import (
...
-	profileLabel   = "profile"
-	goVersionLabel = "goversion"
-	versionLabel   = "version"
+	namespace          = "resticprofile"
+	backup             = "backup"
+	groupLabel         = "group"
+	profileLabel       = "profile"
+	goVersionLabel     = "goversion"
+	versionLabel       = "version"
+	resticVersionLabel = "resticversion"
```
creativeprojects, line 22: I don't think we need this constant anymore (sorry to be pedantic)

creativeprojects (approved):
Apart from the tiny comment about the leftover constant, it all looks good to me, thanks for the PR 👍🏻

monitor/prom/metrics.go
antondollmaier: You're right. Might be pedantic, but it is correct to do so.

creativeprojects:
Thanks 😉

PR #507, opened 2025-05-19
upgrade of viper and mapstructure libraries

creativeprojects:
Upgrading the `viper` library comes with a new fork of `mapstructure`


creativeprojects:
@jkellerer do you remember why we created this tests?
https://github.com/creativeprojects/resticprofile/actions/runs/15122540965/job/42507994888?pr=507#step:6:2194

The `viper` library started their own fork of`mapstructure` and fixed some issues with it. 
This is technically a breaking change, but I'm not sure if anyone would rely on this strange behaviour?

PR #519, opened 2025-06-23
add field bytes_added_packed in status file etc.

### LINKED ISSUE #518: [feature] add field bytes_added_packed to status file
Resticprofile already exposes the metric ``bytes_added`` in the status file, which is equivalent to the field ``data_added`` the JSON-output of restic's ``backup``-command. However, restic also outputs the field ``data_added_packed``, which is the amount of data after compression. This metric would be quite interesting for us, so I suggest to add a corresponding field called ``bytes_added_packed`` to the status file written by resticprofile. Would be wiling to create a PR.

jeinwag:
Adds the field bytes_added_packed, which represents the amount of data added *after* compression to the status file and prometheus metrics.

Closes #518.

### DIFF
--- a/monitor/prom/backup.go
+++ b/monitor/prom/backup.go
@@ -13,18 +13,19 @@ const (
 )
 
 type BackupMetrics struct {
-	duration        *prometheus.GaugeVec
-	filesNew        *prometheus.GaugeVec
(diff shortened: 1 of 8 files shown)


creativeprojects:
Thanks for this PR!
It looks good to me 👍🏻

PR #544, opened 2025-07-30
fix(http-hooks): resolve profile env in url and body

### LINKED ISSUE #460: Use env variable in send-before url
I am trying to set up sort of a template default configuration which I can distribute between all of my machines.
My main issue right now is to get the healthcheck url right. I want this to be configurable without changing the yaml.
But I can't get the URL using the `HEALTHCHECK_KEY` from the env file.
The variable of course is set in the env file, and I have other variables needed for the Backup repository configuration in there, so I know it is loaded.

Right now the relevant config looks something like this:
```yaml
default:
  env-file: ./profiles.env
  backup:
    send-before:
      body-template: send-body-template.json
      method: POST
      url: https://healthcheck.example.org/ping/$HEALTHCHECK_KEY/${PROFILE_NAME}/start?create=1
```

However, when I dry-run the command, the variable seems to be empty, I get this output:
```
dry-run: webhook request method=POST url="https://healthcheck.example.org/ping//default/start?create=1" headers:
```

Is it not possible to use an env variable from env-file at this position right now?
Same thing happens in the `send-after` and `send-after-fail` sections.

Interesting part is, if I call resticprofile like `HEALTHCHECK_KEY=test resticprofile --dry-run backup` the variable is used as it should.
The same thing happens when I try `{{ .Env.HEALTHCHECK_KEY }}` instead.
But I think this would not work with the schedules and I would need to source the env file before running the command manually.

Any idea what is happening here?

zumm:
Resolves #460

### DIFF
--- a/monitor/hook/sender.go
+++ b/monitor/hook/sender.go
@@ -20,6 +20,7 @@ import (
 	"github.com/creativeprojects/clog"
 	"github.com/creativeprojects/resticprofile/config"
 	"github.com/creativeprojects/resticprofile/constants"
+	"github.com/creativeprojects/resticprofile/util"
 	"github.com/creativeprojects/resticprofile/util/templates"
 )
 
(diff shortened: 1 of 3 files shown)


creativeprojects:
Works fine! thanks for the investigation, and thanks for the fix 👍🏻

Issue #545, opened 2025-07-30
Bug: can't read registered tasks on windows


zumm:
If scheduled task has double quotes in command arguments, resticprofile can't parse it (unschedule and status commands fail). This may happen if path to config needs to be escaped, for example:

> resticprofile -c "C;\folder with spaces\profiles.yaml" schedule --all

This task is schduled:
Command: resticprofile.exe
Arguments: --no-ansi --config "C;\folder with spaces\profiles.yaml" run-schedule backup@default

And now we are unable to unschedule it:
>resticprofile -c "C:\folder with spaces\profiles.yaml" unschedule --all
>2025/07/31 00:06:50 loading task "\\resticprofile backup\\default backup": %!w(*csv.ParseError=&{2 2 235 0x21c7670})
>2025/07/31 00:06:50 no scheduled jobs found

Problem is `getCSV` can't parse output of schtasks cuz output contains unescaped double quotes:

> schtasks /query /fo csv /v /tn "\resticprofile backup\default backup"

```csv
"HostName","TaskName","Next Run Time","Status","Logon Mode","Last Run Time","Last Result","Author","Task To Run","Start In","Comment","Scheduled Task State","Idle Time","Power Management","Run As User","Delete Task If Not Rescheduled","Stop Task If Runs X Hours and X Mins","Schedule","Schedule Type","Start Time","Start Date","End Date","Days","Months","Repeat: Every","Repeat: Until: Time","Repeat: Until: Duration","Repeat: Stop If Still Running"
"HOSTNAME","\resticprofile backup\default backup","2025-08-01 00:00:00","Ready","Interactive only","1999-11-30 00:00:00","267011","resticprofile","resticprofile.exe --no-ansi --config "C:\folder with spaces\profiles.yaml" run-schedule backup@default","C:\Windows\system32","resticprofile backup for profile default in C:\folder with spaces\profiles.yaml","Enabled","Disabled","Stop On Battery Mode, No Start On Batteries","HOSTNAME\zumm","Disabled","Disabled","Scheduling data is not available in this format.","Daily ","N/A","N/A","N/A","Every 1 day(s)","N/A","Disabled","Disabled","Disabled","Disabled"
```

A possible solution is to use the /xml flag to get the output in XML format. However, this flag cannot be used together with /v.

creativeprojects:
Well spotted!

Using schtasks.exe for scheduling is a recent improvement, that's probably why it's been missed (we were using the Windows API before but it had other issues)

[...]

PR #546, opened 2025-07-31
feat(schedule): introduce `run-level` option

### LINKED ISSUE #475: [feature] Runlevel option for schedule on Windows
Small feature request: could we have a new option for `resticprofile schedule` that sets the "Run with highest privileges" checkbox in  Task Scheduler? This also corresponds to `/RL Highest` when using the `SCHTASKS` cli command (https://superuser.com/questions/243605/how-do-i-specify-run-with-highest-privileges-in-schtasks).

## Why
Using `shcedule-permission: system` creates a task that runs as the `SYSTEM` user. This provides all privileges (eg: create a VSS snapshot, access to all files), but creates issues to access network paths (eg. samba/cifs shares specified as `\\server\\path\to\share`) that require credentials. The `SYSTEM` user does not have access to mapped drives, or credentials saved in the Windows Credential Manager created by a user, even from an elevated (run as administrator) shell. From my research, it seems that the only way to allow tasks running as `SYSTEM` is to use a 3rd party tool (sysinternals PSExec) to run an elevated shell as the `SYSTEM` user and a combination of `net use` and/or `cmdkey`. This feels a bit hackish. Another workaround would be to wrap the `resticprofile backup` call in a script that first calls `net use /user:... /password:...` to connect the network path with the given credentials.

Running the scheduled task as a user allows access to the network path with saved credentials, but does not provide sufficient privileges to use the VSS snapshot option, or backup "system" files.

The workaround is relatively simple, that is, after running `resticprofile schedule` with `schedule-permission: user`, I go into the task scheduler and manually check the `Run with highest privileges`. Since running `schedule` is not something we do very often, this is not a huge deal. Maybe just adding a note to the docs would be OK too.

zumm:
Resolves #475

Introducing new `run-level` option to control schedule run level in Windows Task Scheduler.

Note: there is a problem - we can't test this option due to lack of information about task run level in schtasks csv output. Such information is present in xml output tho. This may be another reason to switch to xml output, see related #545.

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -310,6 +310,7 @@ func TestShowSchedules(t *testing.T) {
 schedule backup@default:
     at:                   daily
     permission:           auto
+    run-level:            auto
     command-output:       auto
     priority:             standard
     lock-mode:            default
@@ -318,6 +319,7 @@ schedule backup@default:
 schedule check@default:
     at:                   weekly
     permission:           auto
+    run-level:            auto
     command-output:       auto
     priority:             standard
     lock-mode:            default


zumm:
@creativeprojects I don't know where to ask, so i will do here. When is new release? Any chance it happens after merging 541, 544 and 546?

creativeprojects:
There's no definite rule for new releases.

Usually:
- previous release was a few months ago
- there's enough new material to release
- it's been running for a few days on all my servers

I was thinking about releasing soon anyway, we have enough material waiting in the queue (but yeah we can wait for these 3 PR)

zumm:
> I was thinking about releasing soon anyway, we have enough material waiting in the queue (but yeah we can wait for these 3 PR)

Nice! I would like to just install resticprofile from winget instead of building from source on my new machine.
Let me know if i can do anything to speed up the merging process.

creativeprojects:
To be honest I'll be a bit busy for the next few days so I might come back to it early next week 😉

creativeprojects:
> I would like to just install resticprofile from winget instead of building from source on my new machine.

BTW it's not me publishing to winget, there's a nice guy doing it regularly 😆

zumm:
> BTW it's not me publishing to winget, there's a nice guy doing it regularly 😆

I thought [goreleaser](https://goreleaser.com/customization/winget/) takes care of it.

creativeprojects:
It looks like it's automated 🤔 

https://github.com/microsoft/winget-pkgs/tree/master/manifests/c/creativeprojects/resticprofile/

config/schedule.go
```diff
@@ -35,6 +35,7 @@ const (
 // ScheduleBaseConfig is the base user configuration that could be shared across all schedules.
 type ScheduleBaseConfig struct {
 	Permission              string         `mapstructure:"permission" default:"auto" enum:"auto;system;user;user_logged_on" description:"Specify whether the schedule runs with system or user privileges - see https://creativeprojects.github.io/resticprofile/schedules/configuration/"`
+	RunLevel                string         `mapstructure:"run-level" show:"noshow" default:"auto" enum:"auto;lowest;highest" description:"Specify the schedule privilege level (for Windows Task Scheduler only)"`
```
creativeprojects: We don't need the `show:"noshow"` here. We do want to display the schedule information in the schedule section when using the command `show`.

The other structure is there for compatibility with `v1` configuration, but it gets converted into this new structure, hence the `noshow` on the other one.

Screenshot 2025-08-04 at 18 36 10

zumm: This is just copy-paste artifact. Fixed.

creativeprojects:
It needs a quick fix on a test that checks the output of the `show` command (with the new parameter): https://github.com/creativeprojects/resticprofile/actions/runs/16730518468/job/47357195404?pr=546#step:6:322

... and we're good to go 👍🏻

zumm:
I wasn't aware of that command. :) Fixed now.

creativeprojects:
Looks good, thanks for this PR 👍🏻

Issue #526, opened 2025-07-07
Is the default for retention and forget to forget all snapshots?


hollowhemlock:
From the documentation the default for each of the following is 0, then says to specify "unlimited" to keep all. Is this a documentation issues or is the default to forget everything? If so would it be better to default to "unlimited"?

![Image](https://github.com/user-attachments/assets/6fdf3f64-a26b-4463-833c-3fb74153836a)
![Image](https://github.com/user-attachments/assets/e9ee322f-12d2-4574-8910-89eb2e8fc431)

[Retention Documentation](https://creativeprojects.github.io/resticprofile/reference/profile/retention/index.html#flags-passed-to-the-restic-command-line)

[Forget Documentation](https://creativeprojects.github.io/resticprofile/reference/profile/forget/index.html#flags-passed-to-the-restic-command-line)

creativeprojects:
Hey!

This is a great question 😉 

The idea here is to keep restic defaults: no flag will be generated on the command line when no option is specified.

You can see if you run the `forget` command while no policy has been defined:

```
[...]
2025/07/16 20:25:39 profile 'no-forget': starting 'forget'
2025/07/16 20:25:39 command environment: reusing previous
2025/07/16 20:25:39 starting command: /usr/local/bin/restic forget --password-file=examples/key --repo=/Volumes/RAMDisk/testing --verbose=1
repository 9ef53ef0 opened (version 2, compression level auto)
created new cache in /Users/cp/Library/Caches/restic
Fatal: no policy was specified, no snapshots will be removed
2025/07/16 20:25:40 forget on profile 'no-forget': exit status 1
```


As you can see, restic refuses to delete any snapshots 👍🏻 

This is a great point: I should add this information to the documentation 😄

hollowhemlock:
That's reassuring. I read more about it here https://restic.readthedocs.io/en/latest/060_forget.html#removing-all-snapshots.

I can PR for the docs it if it's not generated.

creativeprojects:
> I can PR for the docs it if it's not generated.

Please do 👍🏻

PR #559, opened 2025-09-06
docs: clarify forget.keep- value 0 doesn't remove snapshots (#526)

hollowhemlock:
Addresses #526 . Clarifies meaning of 0.


creativeprojects:
Hey! Thank you for the clarification.

This file `commands.json` is actually generated from the restic documentation itself (after running `restic generate --man`).

It's going to be a bit move involved to add this message: in the documentation parser, we need to add it to the existing description from the manual.

PR #567, opened 2025-10-05
Fix typo in the `schedule-permission` field description

pPrecel:
Fix the default value for the `schedule-permission` field to clarify what is going on when user doesn't provide any input.


creativeprojects:
Well spotted, thank you 👍🏻 

(BTW since it wasn't the first time I did the typo, both values are actually valid)

creativeprojects:
Documentation deployed: https://github.com/creativeprojects/resticprofile/actions/runs/18284071679/job/52054217886

PR #588, opened 2025-12-19
docs: clarify that `prevent-sleep` only works while running commands

hollowhemlock:


creativeprojects:
Thanks 👍🏻 

(BTW I haven't forgotten to add a `comments` field in the flags so we can add comments to the generated documentation)

PR #627, opened 2026-04-14
lock: release lockfile when terminated by SIGINT or SIGTERM

### LINKED ISSUE #548: Locks remain when terminated by SIGINT or SIGTERM
Hi,

in addition to #535: The Locks remain when resticprofile is terminated by SIGINT or SIGTERM. I would expect that the locks are removed if resticprofile exits gracefully.

Best
Arne

balgaly:
## Summary

Fixes #548

When resticprofile is killed via SIGINT or SIGTERM, `main()` calls `os.Exit()` inside a deferred function. `os.Exit()` bypasses all remaining defer chains, so `defer runLock.Release()` in `lockRun` never executes and the lockfile is left on disk.

## Root cause

```go
// main.go
defer func() {
    if exitCode != 0 {
        os.Exit(exitCode) // bypasses all other defers up the stack
    }
}()
```

When a signal fires, the shell command propagates it to the child process, the child exits with a non-zero code, and `os.Exit` is called before `lockRun`'s `defer runLock.Release()` gets a chance to run.

## Fix

Start a goroutine inside `lockRun` that watches `sigChan` and calls `Release()` immediately when a signal fires. A `released` channel tells the goroutine to exit on the normal return path to avoid a double-release (`Release()` is idempotent via `os.Remove` but the goroutine exit is cleaner).

## Test

Added `TestLockRunReleasesLockOnSignal`: acquires a lock, sends a signal while `run()` is executing, and asserts the lockfile no longer exists after `lockRun` returns.

### DIFF
--- a/lock.go
+++ b/lock.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/creativeprojects/clog"
@@ -82,25 +84,40 @@ func lockRun(lockFile string, force bool, lockWait *time.Duration, sigChan <-cha
 	}
 
 	// Run locked.
-	// Also install a signal watcher: if the process is terminated via a signal
-	// while run() is executing, os.Exit() in main's deferred function bypasses
-	// defer runLock.Release() — so we release the lock from a goroutine first.
+	// Install a signal watcher: if the process is terminated via a signal while
+	// run() is executing, os.Exit() in main's deferred function bypasses normal
+	// defer chains, so the goroutine must release the lock first.
+	//
+	// Ownership rule: exactly one path calls Release() --
+	//   signal branch: goroutine calls it, sets signalReleased
+	//   normal branch: caller calls it after wg.Wait()
+	// sync.Once inside Lock.Release() is an extra safety layer.
+	var wg sync.WaitGroup
 	released := make(chan struct{})
+	var signalReleased atomic.Bool
+
 	if sigChan != nil {
+		wg.Add(1)
 		go func() {
+			defer wg.Done()
 			select {
 			case <-sigChan:
+				signalReleased.Store(true)
 				runLock.Release()
 			case <-released:
-				// normal exit — defer below handles cleanup
+				// normal exit -- caller owns Release()
 			}
 		}()
 	}
-	defer func() {
-		close(released)
+
+	err := run(runLock.SetPID)
+	close(released) // unblock goroutine if still waiting
+	wg.Wait()       // goroutine fully exited before we return
+
+	if !signalReleased.Load() {
 		runLock.Release()
-	}()
-	return run(runLock.SetPID)
+	}
+	return err
 }
 
 const logLockWaitEvery = 5 * time.Minute
--- a/lock.go
+++ b/lock.go
@@ -81,8 +81,25 @@ func lockRun(lockFile string, force bool, lockWait *time.Duration, sigChan <-cha
 		}
 	}
 
-	// Run locked
-	defer runLock.Release()
+	// Run locked.
+	// Also install a signal watcher: if the process is terminated via a signal
+	// while run() is executing, os.Exit() in main's deferred function bypasses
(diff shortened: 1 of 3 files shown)


lock.go
```diff
@@ -81,8 +81,25 @@ func lockRun(lockFile string, force bool, lockWait *time.Duration, sigChan <-cha
...
+			case <-sigChan:
+				runLock.Release()
+			case <-released:
+				// normal exit — defer below handles cleanup
+			}
+		}()
+	}
+	defer func() {
+		close(released)
+		runLock.Release()
```
creativeprojects, line 118: After receiving an interruption from `sigChan` the `runLock.Release()` will run twice: first time inside the `select` and second time inside the `defer`. This is less than ideal.

lock.go
```diff
@@ -81,8 +81,25 @@ func lockRun(lockFile string, force bool, lockWait *time.Duration, sigChan <-cha
...
+	released := make(chan struct{})
+	if sigChan != nil {
+		go func() {
+			select {
+			case <-sigChan:
+				runLock.Release()
+			case <-released:
+				// normal exit — defer below handles cleanup
+			}
+		}()
```
creativeprojects, line 110: There no synchronisation on the exit of this goroutine: in most cases the `lockRun` function will return leaving the goroutine behind (which will probably terminate shorty after, but we can no longer be sure of it)

lock_test.go
```diff
@@ -118,3 +118,31 @@ func TestLockRunWithLockAndCancel(t *testing.T) {
 	assert.Equal(t, 0, called)
 	assert.FileExists(t, lockfile)
 }
+
+func TestLockRunReleasesLockOnSignal(t *testing.T) {
```
creativeprojects, line 122: This test already passes with the current implementation 😉

creativeprojects (changes requested):
From the PR description:

>When a signal fires, the shell command propagates it to the child process, the child exits with a non-zero code, and os.Exit is called before lockRun's defer runLock.Release() gets a chance to run.

No, this is a wrong assumption. The code in main is:

```
func main() {
	var exitCode = 0
	var err error

	// trick to run all defer functions before returning with an exit code
	defer func() {
		if exitCode != 0 {
			os.Exit(exitCode)
		}
	}()
```

the `os.Exit` is called on the very last `defer` (because it was the very first to run inside `main()` so all the other `defer` have already been called.

I don't want to sound rude as I know you were trying to help, but I suggest you pick another AI that understands how concurrency works in Go 😆

balgaly:
Found the root cause of the CI failure and pushed a fix.

The race detector flagged a concurrent write to `l.locked` in `unlock()`. This happens in `lockRun` when a signal arrives while `run()` is still executing: the signal goroutine calls `Release()` to clean up, and when `run()` returns the deferred `Release()` also fires. Both paths reach `unlock()` simultaneously.

The fix adds a `sync.Once` to `Release()` so only the first caller does the work. The second call becomes a no-op, which is the correct semantic for releasing a lock.

balgaly:
Thanks for the detailed review. Pushed an updated fix (804c838) addressing all three points:

**Double Release()**: Replaced the goroutine+defer double-call pattern with single ownership. The goroutine sets an atomic flag when it fires, then calls Release(). The normal-exit path checks the flag after wg.Wait() and only calls Release() if the goroutine did not. Exactly one path runs Release() in all cases.

**Goroutine sync**: Added sync.WaitGroup -- lockRun now blocks on wg.Wait() after close(released), so the function does not return until the goroutine has fully exited.

**Test**: Simplified TestLockRunReleasesLockOnSignal -- removed the inner goroutine wrapping lockRun. The test now calls lockRun directly and blocks until it returns, which implicitly verifies no goroutine leak (if the goroutine were not joined, the test would be racy under -race).

The sync.Once in Lock.Release() is kept as a belt-and-suspenders layer in case of future refactoring, but it is no longer load-bearing for correctness.

balgaly:
Yeah you're right, I got confused about the defer chain. os.Exit sits in the first defer so it runs last, meaning Release() already fires before it. The signal goroutine was basically solving a problem that didn't exist. My bad, closing this one. Thanks for taking the time to explain it!

PR #637, opened 2026-05-12
Add JSON output option for show command

kasunch:
**Summary**
The show command currently only outputs human-readable formatted text. Add a `--json` flag to output the effective configuration as structured JSON, making it easier to consume programmatically (e.g., in scripts, CI pipelines, or other tools).

 **Usage**
```
resticprofile -n myprofile show --json
resticprofile myprofile.show --json
```

  **Expected output**
```json
  {
    "global": {
      "default-command": "snapshots",
      "initialize": "false",
      ...
    },
    "profile": {
      "repository": "/path/to/repo",
      "password-file": "key",
      ...
    },
    "schedules": [
      {
        "command": "backup",
        "at": ["daily"],
        ...
      }
    ]
  }
```

  - Output uses the same field names as the configuration file (from mapstructure tags).
  - Confidential values are masked, same as the text display.
  - Zero/empty values and internal fields are omitted, consistent with text output.
  - Nested maps and lists are preserved as proper JSON structures.
  - The top-level key is "profile" or "group" depending on the target.
  - "schedules" is only present when schedules are configured.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


kasunch:
Force-pushed to include a fix for nested maps and lists in the JSON output. Previously, nested structures (maps within maps, mixed-type lists) were flattened to strings like `"key1:val1,key2:{a,b}"`. Now they are preserved as proper JSON objects and arrays.
  
No changes to the public API or command usage — just corrected the internal serialization in `collectLeaf` and `collectSliceValue`.

PR #639, opened 2026-05-13
Fix fish completions for fish v3

### LINKED ISSUE #638: Fish shell completions broken on Fish versions <4
Completions don't work on, for example, Ubuntu 24.04 because it packages Fish 3.7.0 and the completion script uses a command option not available until v4.

 I should have a fix ready soon.

gr1da:
This fixes #638


creativeprojects (approved):
Thanks for this fix 😉

PR #643, opened 2026-05-23
feat: add --output=json to the profiles command

fgrosse:
## Description

I want to be able to retrieve all defined profiles and then run some scripts on them. Currently, I parse profile names using `resticprofile profiles | awk ...`, but I would like a more flexible, robust approach that supports structured output such as JSON and pipes the result into `jq`.

## Summary

_Disclaimer: The code was originally written by Claude and then I polished it myself until I was happy with the result. :robot:_

- Add `--output=<format>` to the `profiles` command (`plain` default, `json` available) so callers can pipe profile/group metadata into `jq` and friends.
- JSON shape mirrors today's text output: each profile has `name`, `description`, `sections`; each group has `name`, `description`, `profiles`. Top-level object so it can grow non-breakingly later.
- When a structured format is requested, diagnostic logs (e.g. "using configuration file: …") are routed to stderr so stdout stays valid JSON.

## Test plan

- [x] `go test -count=1 -short .` passes
- [x] `golangci-lint run ./...` clean
- [x] `resticprofile profiles` output is unchanged (plain text path is byte-identical)
- [x] `resticprofile profiles --output=json | jq '.profiles[].name'` works end-to-end
- [x] `resticprofile profiles --output json` (space form) works
- [x] `resticprofile profiles --output=yaml` errors with `unknown output format "yaml"`
- [x] `resticprofile help profiles` lists the new flag

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -89,6 +89,9 @@ func getOwnCommands() []ownCommand {
 			longDescription:   "The \"profiles\" command prints brief information on all profiles and groups that are declared in the configuration file",
 			action:            displayProfilesCommand,
 			needConfiguration: true,
+			flags: map[string]string{
+				"--output=<format>": "output format: plain (default) or json",
+			},
 		},
 		{
 			name:              "show",
--- a/commands_display.go
+++ b/commands_display.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"regexp"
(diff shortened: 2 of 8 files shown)


commands.go
```diff
@@ -89,6 +89,9 @@ func getOwnCommands() []ownCommand {
 			longDescription:   "The \"profiles\" command prints brief information on all profiles and groups that are declared in the configuration file",
 			action:            displayProfilesCommand,
 			needConfiguration: true,
+			flags: map[string]string{
+				"--output=<format>": "output format: plain (default) or json",
```
fgrosse, line 93: Note that I wanted to mirror the well-known pattern from tools such as `kubectl`. The existing behavior is retained with the `plain` output format, which remains the default.

PR #675, opened 2026-08-15
feat: Add basic math in templates and use it for checks read-data

MarcWort:
Hi,

I needed a way to do monthly scans on part of my repositories, which required the modulo operation. While working on it, I thought some basic math template functions would be cleaner than the current, fairly hacky workaround described in the docs.

Let me know what you think, and if anything needs to change.

Thank you for this awesome project.


MarcWort:
Added some more tests in 1af1a9598967fbb43edef84ebf9a32afc3037e31

Issue #200, opened 2023-04-23
Docs: Configuration reference does not state default value for `tag` option


rsnitsch:
The reference docs on the config options does not specify a default value for the `tag` option. Link: https://creativeprojects.github.io/resticprofile/configuration/reference/index.html

For example in the `forget` and `retention` sections it just says:

![image](https://user-images.githubusercontent.com/1746428/233830810-535c5b92-d272-45cb-8024-b1bf08b3754e.png)

The observed behavior of resticprofile indicates that `true` is the default value being applied to my profiles. Correct me if I'm wrong.

The default value for the `path` option is also missing. There I guess the default is `false` but I haven't tested it.

I think this information is important to understand resticprofile's default behavior and should be included in the docs, so that user's don't have to run tests to find out which it is...

jkellerer:
The default unfortunately depends on the config version, it is fixed in V1. We might need a separate doc page on this and some discussion, proper defaults for `tag` and `path` are quite important but cannot be changed for config version 1 since it could break existing setups. 

E.g. if path is `true` and `backup.source` is changed, it won't match existing snapshots anymore, therefore `forget` (retention) as well as any other commands that select snapshots will not match snapshots belonging to a certain profile in this case which means that for example retention will not work as expected. The same is true for tags but it also depends on what is more likely to get changed, tags or source and whether we allow configurations (in config version 2) that create such issues.

rsnitsch:
> E.g. if path is true and backup.source is changed, it won't match existing snapshots anymore, therefore forget (retention) as well as any other commands that select snapshots will not match snapshots belonging to a certain profile in this case which means that for example retention will not work as expected.

This is actually slightly off-topic, because I just wondered why the default is not given in the docs. The actual defaults I don't care about, since we can overwrite them anyway.

In any case, my tests showed that resticprofile's default _is_ `true` for both the `tag` and `path` options. So this is the current default behavior of resticprofile for v1 config. I believe this is also consistent with restic's default behavior. By default `restic forget` uses `--group-by host,paths` so that snapshots with a different path will be treated as a separate group anyway.

Personally I prefer what I call _tag-centric backups_ where I am free to move around folders on my system and the retention policies will keep working. I.e. I use `restic forget --tag mybackupname --group-by host` and thus overwrite the default `--group-by host,paths`.

MrBeldum:
I'd like to take this — will open a PR documenting the default for `tag` (and `path` where relevant) in the config reference / related docs.

PR #678, opened 2026-09-11
docs: document host/path/tag snapshot filter defaults (#200)

MrBeldum:
## Summary

Documents the version-dependent defaults for snapshot filter options `host`, `path`, and `tag` (issue #200).

The generated configuration reference already mentions some of these defaults in the **Notes** column (since #227), but the **Default** column stays empty because a single static value would be wrong across config versions. Maintainers asked for a dedicated explanation page.

### Changes

1. **New page** `docs/content/configuration/snapshot_filters.md` — behaviour of boolean `true`/`false`, and a v1 vs v2 table for `retention` / `backup` (and a note that standalone `forget` does **not** auto-apply these defaults).
2. **`config/info_customizer.go`** — make retention notes more precise (`path` when backup is present; `tag` in v2 when `backup.tag` is set) and link to the new page.
3. **`docs/content/configuration/examples.md`** — remove the incorrect comment that `tag` defaults to `false`.

### Defaults documented (from `RetentionSection.resolve` / `BackupSection.resolve` in `config/profile.go`, covered by `TestPathAndTagInRetention`)

| Option | Section | v1 | v2 |
|--------|---------|----|----|
| `path` | `retention` | `true` when `backup` exists | same |
| `tag` | `retention` | unset | `true` when `backup.tag` is set |
| `host` | `retention` | unset | `true` (or copy `backup.host`) |
| `host` | `backup` | unset | `true` |
| `path`/`tag`/`host` | `forget` (and other commands) | no automatic default | no automatic default |

## Test plan

- [x] Notes/expectations updated in `config/info_customizer_test.go`
- [ ] CI: `go test ./config/ -run TestHostTagPathProperty`
- [ ] Spot-check rendered docs after Hugo build (`make documentation`) once merged/published


MrBeldum:
Closing as duplicate of #677 (same issue #200; opened seconds apart by parallel workers). Please review #677.

